PR #1, opened 2020-10-08
dummy file

Maquiner:


pipeline/eBOSS_QSO/QuasarMapper.py
```diff
@@ -0,0 +1,191 @@
+import matplotlib.pyplot as plt
```
damonge, line 1: No plotting

pipeline/eBOSS_QSO/QuasarMapper.py
```diff
@@ -0,0 +1,191 @@
...
+from astropy.io import fits
+from astropy.table import Table
+
+import os
+import pandas as pd
+import collections
+import numpy as np
+import healpy as hp
+
+import pymaster as nmt
```
damonge, line 11: You won't need this

pipeline/eBOSS_QSO/QuasarMapper.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class QuasarMapper:
+    def __init__(self):
```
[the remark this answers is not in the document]
damonge: So the signature of the init function should be:
```
def __init__(self, data_list, random_list):
```
where `data_list` and `random_list` are a two lists of strings, and each string corresponds to the path to a fits file corresponding a FITS catalog that will be included in the final map.

pipeline/eBOSS_QSO/QuasarMapper.py
```diff
@@ -0,0 +1,191 @@
...
+import healpy as hp
+
+import pymaster as nmt
+
+
+class QuasarMapper:
+    def __init__(self): 
+        #Inputs: the four fits files from eBOSS DR14
+        
+        data_path = '/home/zcapjru/PhD/Data/'
```
damonge, line 18: Now, all of these hard-coded things you can replace by more generic looping over the input file paths:
```
self.cat_data = []
self.cat_random = []
for file_data, file_random in zip(data_list, random_list):
    if not os.path.isfile(file_data):
        raise ValueError(f"File {file_data} not found")
    with fits.open(file_data) as f:
        self.cat_data.append(Table.read(f).to_pandas()
    if not os.path.isfile(file_random):
        raise ValueError(f"File {file_random} not found")
    with fits.open(file_random) as f:
        self.cat_random.append(Table.read(f).to_pandas()
```
At this point, `self.cat_data` and `self.cat_random` will be lists of catalog tables containing data and randoms.

pipeline/eBOSS_QSO/QuasarMapper.py
```diff
@@ -0,0 +1,191 @@
...
+            print('missing NGC random data')
+           
+        if os.path.exists(data_path+'eBOSS_QSO_clustering_random-SGC-vDR16.fits') :
+            print('SGC random data found')
+            self.SGC_r_data =  fits.open(data_path + 'eBOSS_QSO_clustering_random-SGC-vDR16.fits')
+        else :
+            print('missing SGC random data')
+        
+
+        self.nside = 512
```
damonge, line 45: Make `nside` another input parameter (instead of hardcoding it). You can give it the default value of 512 if you want.

pipeline/eBOSS_QSO/QuasarMapper.py
```diff
@@ -0,0 +1,191 @@
...
+        #set resolution
+        
+        self.bands = nmt.NmtBin.from_nside_linear(self.nside, 20)
+        #set l's per band
+        
+        self.NGC = Table.read(self.NGC_data).to_pandas()
```
damonge, line 51: All this is superseded by what I wrote above

pipeline/eBOSS_QSO/QuasarMapper.py
```diff
@@ -0,0 +1,191 @@
...
+        self.NGC_r = Table.read(self.NGC_r_data).to_pandas()
+        self.SGC = Table.read(self.SGC_data).to_pandas()
+        self.SGC_r = Table.read(self.SGC_r_data).to_pandas()
+        #from .fits to table to managable pandas frames
+
+        #Merge NGC + SGC
+        self.edges = [1.5] #inner bins edges
+        self.whole =  pd.concat([self.NGC, self.SGC])
```
damonge, line 59: Here you could do `self.whole = pd.concat(self.cat_data)`. Doing so you will be storing things twice, so you could even overwrite `cat_data` (i.e. `self.cat_data = pd.concat(self.cat_data)`).

Do the same with `self.whole_r` below.

pipeline/eBOSS_QSO/QuasarMapper.py
```diff
@@ -0,0 +1,191 @@
...
+        for i in range(len(self.edges)+1):
+            self.nmean_field  =  self.get_nmean_map(self.binned[i])
+            self.nmean_random =  self.get_nmean_map(self.binned_r[i])
+            self.alpha = self.get_alpha(self.nmean_field, self.nmean_random)
+            
+            self.maps["bin_{}".format(i)+"_delta"]= self.get_delta_map(self.nmean_field, self.nmean_random, self.alpha) 
+            self.maps["bin_{}".format(i)+"_nmean"]= self.nmean_random*self.alpha
+            self.maps["bin_{}".format(i)+"_mask"] = self.nmean_random
+               
+        print('Maps computed succesfully')
```
damonge, line 76: No printing (or, if you do, enclose in something like `if self.verbose`, or something like that).

pipeline/eBOSS_QSO/QuasarMapper.py
```diff
@@ -0,0 +1,191 @@
...
+        self.whole =  pd.concat([self.NGC, self.SGC])
+        self.whole_r =  pd.concat([self.NGC_r, self.SGC_r])
+        
+        self.binned   = self.bin_z(self.whole ,  self.edges)
+        self.binned_r = self.bin_z(self.whole_r ,  self.edges)
+        #arrays containing the binned data for each field
+        
+        self.maps = {}
+        for i in range(len(self.edges)+1):
+            self.nmean_field  =  self.get_nmean_map(self.binned[i])
```
damonge, line 68: This is good, but see my comment below about making all of these private methods.

pipeline/eBOSS_QSO/QuasarMapper.py
```diff
@@ -0,0 +1,191 @@
...
+    
+    def get_alpha(self, nmean_field, nmean_random):
+        #imputs: pandas frames
+        #Calculates the mean quasar count per pixel field to random ratio
+        
+        alpha = sum(nmean_field)/sum(nmean_random)
+
+        return alpha
+    
+    def get_delta_map(self, nmean_field, nmean_random, alpha):
```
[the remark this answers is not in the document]
damonge: In your case, `nmean` and `mask` are the same, but let's make it more general.

pipeline/eBOSS_QSO/QuasarMapper.py
```diff
@@ -0,0 +1,191 @@
...
+        #David's notes formula
+        N_ell = pixel_A*np.mean(sum_items)
+        
+        nl_coupled = np.array([N_ell * np.ones(3*self.nside)])
+        nl_decouple = wsp.decouple_cell(nl_coupled) #removed bias here
+
+        #Following Carlos code
+        return nl_decouple
+
+output = QuasarMapper()
```
damonge, line 183: We will want this to be a python module that doesn't execute any code. So I'd say remove this, and test the code by importing it into a different script.

pipeline/eBOSS_QSO/QuasarMapper.py
```diff
@@ -0,0 +1,191 @@
...
+
+f = open("cls.txt","w")
```
damonge, line 185: You won't need this here, but FYI, the easy and nice way of doing this is using [numpy.savetxt](https://docs.scipy.org/doc/numpy/reference/generated/numpy.savetxt.html) (and [numpy.loadtxt](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.loadtxt.html) for reading).

damonge (commented):
Great, @Maquiner . This does what we want, but we want this class to be more general (i.e. not tied up to this particular quasar catalog). See the comments below.

PR #4, opened 2020-12-18
Jaime xcell

damonge:


carlosggarcia (approved):
We revised the changes in our weekly meeting and Jaime fixed the last errors. DESwl and KV450 still need testing (but they needed it before, too).

PR #18, opened 2021-02-10
added test for cov + option to vary fsky in dummy mapper

carlosggarcia:
Let me know if you think of other tests we can add for the covariance.


damonge (approved):
lgtm!

PR #24, opened 2021-02-11
Kv e2

Maquiner:
Just changed the convention of KV


xcell/mappers/mapper_DESY1wl.py
```diff
@@ -108,10 +117,43 @@ def _set_mode(self, mode=None):
             raise ValueError(f"Unknown mode {mode}")
         return e1_flag, e2_flag, mode
 
+    def _get_Rs():
```
carlosggarcia, line 120: I'd check that Rs is not already set. Otherwise, if it is somehow called, it will fail as you will have had clean the catalog.

xcell/mappers/mapper_DESY1wl.py
```diff
@@ -33,7 +33,13 @@ def __init__(self, config):
         self.nz = None
         # load cat
         self.cat_data = self._load_catalog()
+        # get items for calibration
+        self._get_Rs()
```
carlosggarcia, line 37: I tend to prefer that `get_something` returns `something`; i.e. I would change that to `self.Rs = self._get_Rs()`

xcell/mappers/mapper_DESY1wl.py
```diff
@@ -67,23 +75,24 @@ def _load_catalog(self):
             cat_zbin = Table.read(self.config['zbin_cat'],
                                   format='fits', memmap=True)
             cat_zbin.keep_columns(columns_zbin)
-            self.cat_data = astropy.table.join(self.cat_data, cat_zbin)
-            # Note: By default join uses checks the values of the column named
-            # the same in both tables
-            col_w = Column(name='weight',
-                           data=np.ones(len(self.cat_data), dtype=int))
-            self.cat_data.add_column(col_w)
+            self.cat_data = np.hstack((self.cat_data, cat_zbin()
```
carlosggarcia, line 78: Use `astropy.table` functions: https://docs.astropy.org/en/stable/api/astropy.table.hstack.html

xcell/mappers/mapper_DESY1wl.py
```diff
@@ -67,23 +75,24 @@ def _load_catalog(self):
...
 
             # remove bins which are not the one of interest
-            self.cat_data.remove_rows(self.cat_data['zbin_mcal'] != self.zbin)
+            # Logic: If item in one of zbins --> sel = False
+            # Thus it is not removed by next line
+            sel = self.cat_data['zbin_mcal'] != self.zbin and \
```
carlosggarcia, line 83: Does this work? Doesn't it complain with `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`? If so, change `and` to `*` (i.e. multiply the arrays).

xcell/mappers/mapper_DESY1wl.py
```diff
@@ -108,10 +117,43 @@ def _set_mode(self, mode=None):
...
+        mean_e1_2p = np.mean(data_2p['e1'])
+        mean_e2_2p = np.mean(data_2p['e2'])
+        mean_e1_2m = np.mean(data_2m['e1']) 
+        mean_e2_2m = np.mean(data_2m['e1'])
+
+        self.Rs = np.array([[(mean_e1_1p-mean_e1_1m)/0.02,
+                             (mean_e1_2p-mean_e2_2m)/0.02],
+                            [(mean_e2_1p-mean_e2_1m)/0.02,
+                             (mean_e2_2p-mean_e2_2m)/0.02]])
+        return
```
carlosggarcia, line 139: As said before, I'd return `Rs` and not set directly `self.Rs` in this function

xcell/mappers/mapper_DESY1wl.py
```diff
@@ -133,7 +175,7 @@ def get_signal_map(self, mode=None):
         self.signal_map = self.maps[mod]
         return self.signal_map
 
-    def get_nz(self):
+    def get_nz(self, dz=0):
```
carlosggarcia, line 178: You have to add the lines that do something with the `dz`. Copy those in `mapper_dummy.py`

xcell/mappers/mapper_KV450.py
```diff
@@ -151,7 +151,7 @@ def get_signal_map(self, mode=None):
             goodpix = mask > 0
             we1[goodpix] /= mask[goodpix]
             we2[goodpix] /= mask[goodpix]
-            self.maps[mod] = [-we1, we2]
+            self.maps[mod] = [we1, -we2]
```
carlosggarcia, line 154: This should be in another branch and in another pull request.

carlosggarcia (changes requested):
Flake!

carlosggarcia:
Also, modify `test_mapper_desy1wl.py` accordingly.

PR #25, opened 2021-02-11
Deswl

Maquiner:


xcell/mappers/mapper_DESY1wl.py
```diff
@@ -108,11 +116,44 @@ def _set_mode(self, mode=None):
             raise ValueError(f"Unknown mode {mode}")
         return e1_flag, e2_flag, mode
 
+    def _get_Rs(self):
+        data_1p = self.cat_data[self.cat_data['zbin_mcal_1p'] == self.zbin]
```
carlosggarcia, line 120: return self.Rs if already computed

xcell/mappers/mapper_DESY1wl.py
```diff
@@ -133,7 +174,7 @@ def get_signal_map(self, mode=None):
         self.signal_map = self.maps[mod]
         return self.signal_map
 
-    def get_nz(self):
+    def get_nz(self, dz=0):
```
carlosggarcia, line 177: Add effect with dz! (I'll do it in a sec)

carlosggarcia (changes requested):

carlosggarcia (approved):

PR #30, opened 2021-02-18
Planck kappa modifications

damonge:


carlosggarcia (approved):
The plots look OK (except for a peak at high ell due to noise)

PR #45, opened 2021-03-03
kv450 sign flip

damonge:
Fixes sign convention for KV450 after computing the xcorr with DECALS


carlosggarcia:
You've got to update the unit test too, to accommodate the sign flip

PR #47, opened 2021-03-04
Decals test

Maquiner:


carlosggarcia (changes requested):
Is `xcell/tests/data/alm.fits` empty? If so, delete it.

damonge (approved):

Issue #46, opened 2021-03-04
Fast-mode on mappers


damonge:
Since sometimes we need to load mappers to access properties such as dndzs, data type, spin, it'd be good for them to have a "fast" flag in their `__init__` that allows you to create a quick instance of them that doesn't need to load any big data files (catalogs, masks etc.)

carlosggarcia:
Instead of adding a new flag `fast` it'd be better to make all mappers fast by default; i.e. avoid loading everything in the `__init__`, just under request when neeed. This is what was done in #52 for DESY1wl

Maquiner:
Solved!

PR #52, opened 2021-03-10
mapper_DESY1wl: Speed up

carlosggarcia:
I have improved the mapper_DESY1wl mapper by two means:
1. Only load the catalog if necessary
2. Save compressed `.fits.gz` maps which are read faster than computing them again and again.

This could be extended to other mappers in case we see they're slow too and would substitute the `fast` flag proposal (issue #46 ); i.e. they should be fast by default.


damonge (approved):
LGTM!

PR #56, opened 2021-03-11
rest of maps

Maquiner:


xcell/mappers/mapper_eBOSSQSO.py
```diff
@@ -21,82 +21,96 @@ def __init__(self, config):
         """
         self._get_defaults(config)
 
-        self.cat_data = []
-        self.cat_random = []
-        self.z_arr_dim = config.get('z_arr_dim', 50)
+        self.cats = {'data': [], 'random': []}
```
damonge, line 24: As with decals, I'd make these `None` and `None` to be consistent with the other mappers.

xcell/mappers/mapper_eBOSSQSO.py
```diff
@@ -21,82 +21,96 @@ def __init__(self, config):
...
         self.nl_coupled = None
         self.mask = None
 
+    def get_catalog(self, mod='data'):
+        if mod == 'data':
+            data_file = self.config['data_catalogs']
+        else:
+            data_file = self.config['random_catalogs']
+
+        if len(self.cats[mod]) == 0:
```
damonge, line 45: And as before:
```
if self.cats[mod] == None:
    self.cats[mod] = []
```

damonge (changes requested):
Do we want to save maps and masks? This is not done for DESY1gc or eBOSS

xcell/mappers/mapper_eBOSSQSO.py
Maquiner: I see! I was struggling with that. Thanks!

Maquiner: resolved!

damonge (approved):
LGTM!

PR #57, opened 2021-03-11
decals_cats

Maquiner:


damonge (changes requested):
Do we want to be consistent and save the signal maps too?

xcell/mappers/mapper_DECaLS.py
```diff
@@ -31,25 +31,16 @@ def __init__(self, config):
         self.cat_data = []
```
damonge, line 31: I would make this `self.cat_data = None` (see comment below)

xcell/mappers/mapper_DECaLS.py
```diff
@@ -58,21 +49,37 @@ def __init__(self, config):
...
 
-    def _get_angmask(self, cat):
-        bmask = hp.read_map(self.config['binary_mask'],
-                            verbose=False)
-        nside = hp.npix2nside(len(bmask))
-        ipix = hp.ang2pix(nside, cat['RA'], cat['DEC'],
-                          lonlat=True)
-        return bmask[ipix] > 0.
+    def get_catalogs(self):
+        if len(self.cat_data) == 0:
```
damonge, line 53: then make this
```
if self.cat_data == None:
    self.cat_data = []
```
Just to be consistent with the other mappers.

damonge (changes requested):

xcell/mappers/mapper_DECaLS.py
Maquiner: resolved!

damonge (approved):
LGTM!

PR #59, opened 2021-03-11
kids fix

Maquiner:
fixed by invoking self. inside method

### DIFF
--- a/xcell/mappers/mapper_KV450.py
+++ b/xcell/mappers/mapper_KV450.py
@@ -46,6 +46,7 @@ class MapperKV450(MapperBase):
         self.cat_data = None
 
         self.w2s2 = None
+        self.w2s2s = {'PSF': None, 'shear': None, 'stars': None}
 
         self.dndz = np.loadtxt(self.config['file_nz'], unpack=True)
         self.sel = {'galaxies': 1, 'stars': 0}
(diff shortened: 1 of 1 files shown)


damonge (changes requested):
Were we gonna save the masks too?

xcell/mappers/mapper_KV450.py
```diff
@@ -200,24 +202,40 @@ def get_signal_map(self, mode=None):
...
             self.masks[kind] = get_map_from_points(data, self.nside,
                                                    w=data['weight'],
                                                    ra_name='ALPHA_J2000',
                                                    dec_name='DELTA_J2000')
+            hp.write_map(fname_lite, self.masks[kind], overwrite=True)
         self.mask = self.masks[kind]
         return self.mask
 
+    def _get_w2s2(self, mode):
+        kind, e1f, e2f, mod = self._set_mode(mode)
```
damonge, line 226: oh, can you do the same thing with this map?

damonge (changes requested):

xcell/mappers/mapper_KV450.py
Maquiner: done!

damonge (approved):
LGTM!

PR #67, opened 2021-04-10
KiDS 1000

damonge:
Implements KiDS 1000 mapper.
Still need to check a couple of details regarding shear calibration.

### DIFF
--- a/xcell/mappers/mapper_KiDS1000.py
+++ b/xcell/mappers/mapper_KiDS1000.py
@@ -1,35 +1,28 @@
 from .mapper_base import MapperBase
 from .utils import get_map_from_points
-from astropy.table import Table, vstack
+from astropy.table import Table
 import numpy as np
 import healpy as hp
 import os
 
 
-class MapperKV450(MapperBase):
+class MapperKiDS1000(MapperBase):
     def __init__(self, config):
         """
         config - dict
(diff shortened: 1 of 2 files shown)


xcell/mappers/mapper_KiDS1000.py
```diff
@@ -0,0 +1,261 @@
...
+            print('Computing bin{} signal map'.format(self.zbin))
+            data = self._get_gals_or_stars(kind)
+            wcol = data['weight']*data[e1f]
+            we1 = get_map_from_points(data, self.nside, w=wcol,
+                                      ra_name='RAJ2000',
+                                      dec_name='DECJ2000')
+            wcol = data['weight']*data[e2f]
+            we2 = get_map_from_points(data, self.nside, w=wcol,
+                                      ra_name='RAJ2000',
+                                      dec_name='DECJ2000')
```
carlosggarcia, line 176: Just to confirm. The old variables `ALPHA_J2000` and `DELTA_J2000` exist in the catalog. E.g. for `ALPHA_J2000`: 

```
SExtractor named Centroid sky position right ascension (J2000) (deg) 
```
Are we sure `RAJ2000` and `DECJ2000` are the right ones?

damonge: They should be the same to an absurd precision. But it's a good point, let me use the other ones so we don't have to change the test files.

carlosggarcia (approved):

PR #106, opened 2021-06-10
DESY1wl tests

### LINKED ISSUE #92: Missing tests: DESY1wl
- [x] Read data from full catalog (instead of lite version)
- [x] Read shear maps from lite versions (i.e. generate them, write them and then read them again)
- [x] Test `get_nz`
- [x] Test `get_dtype`

carlosggarcia:
Closes #92

Missing tests of reading full catalog and save/read lite catalog


xcell/tests/test_mapper_desy1wl.py
```diff
@@ -29,23 +37,73 @@ def get_es():
...
-    xc.mappers.MapperDESY1wl(config)
+    m_lite = xc.mappers.MapperDESY1wl(config)
+    cat_from_lite = m_lite.get_catalog()
+    s_from_lite = m_lite.get_signal_map()
+    psf_from_lite = m_lite.get_signal_map('PSF')
+    mask_from_lite = m_lite.get_mask()
+    assert np.all(cat == cat_from_lite)
+    assert np.all(np.array(s) == np.array(s_from_lite))
+    assert np.all(np.array(psf) == np.array(psf_from_lite))
+    assert np.all(mask == mask_from_lite)
```
damonge, line 103: Can you add the coupled nl to this list? That way we check that the `w2s2` map is also read from lite.

damonge (changes requested):
Just one small thing. Everything else looks great.

xcell/tests/test_mapper_desy1wl.py
carlosggarcia: Good catch. Done.

damonge (approved):
lgtm

PR #116, opened 2021-06-14
move pip cache above

carlosggarcia:
I think the cache of pip is doing nothing because it's done after pip installation. Trying to move it before

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,12 @@ jobs:
           uses: actions/setup-python@v2
           with:
             python-version: ${{ matrix.py }}
+                    
+        - name: Cache pip
+          uses: actions/cache@v2
+          with:
+            path: ${{ env.pythonLocation }}
+            key: ${{ env.pythonLocation }}-cache
 
         - name: Install GSL, FFTW (linux)
           run: |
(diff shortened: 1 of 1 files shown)


carlosggarcia:
Now it's using cache but the slowest part is running the installation of NaMaster, CCL and CAMB. Isn't it possible to cache those too? (the downloaded files are cached, but not the installation)

damonge:
FYI namaster does not have wheels

carlosggarcia:
Are you sure? I think pip creates it. See line 114 of this action https://github.com/xC-ell/xCell/runs/2827608349?check_suite_focus=true#step:6:114

damonge:
sorry, I meant they're not on pypi, so they need to be built every time you install it.

carlosggarcia:
But the idea with having the cache is avoiding reinstalling again unless the version changes, right?

damonge:
sure, sure, just letting you know in case it plays a role

carlosggarcia:
Got it! Based on https://medium.com/ai2-blog/python-caching-in-github-actions-e9452698e98d

carlosggarcia:
Closing this PR and opening a new one after squashing commits

carlosggarcia:
Reopen because I saw that there's only 1 commit that will be merged

damonge (approved):
LGTM

PR #134, opened 2021-07-28
Halo model theory

### LINKED ISSUE #132: Halo model for covariance
Some maps (e.g. tSZ maps, but also the 2MPZ and WIxSC maps) would benefit from having their associated theory C_ells calculated using the halo model (e.g. rather than assuming a linear bias for the galaxy clustering tracers). Incorporating some aspects of the halo model in the code will also allow us to include non-Gaussian predictions for the covariance further down the line.

### LINKED ISSUE #133: gc_bias
@carlosggarcia  what's the point of `'gc_bias'` in the `fiducial` part of the param file?
See: https://github.com/xC-ell/xCell/blob/069c42389f56dfff3a209eef4d05175707c98744/xcell/cls/cl.py#L246

Wouldn't it be easier to make this just:
```
bias = (z, tracer.get('bias', 1) * np.ones_like(z))
```

damonge:
Closes #133 
Closes #132


damonge:
Forgot to make this a draft PR. This is not yet ready for review.

xcell/cls/cl.py
```diff
@@ -232,9 +232,52 @@ def __init__(self, data, tr1, tr2):
         self.outdir = self.get_outdir('fiducial')
         os.makedirs(self.outdir, exist_ok=True)
         self.cosmo = self.get_cosmo_ccl()
+        self.hm_par = self._get_halomodel_params()
         self._ccl_tr1 = None
         self._ccl_tr2 = None
 
+    def _get_halomodel_params(self):
+        if 'halo_model' not in self.data.data['cov']:
+            self.data.data['cov']['halo_model'] = {}
```
carlosggarcia, line 241: What would you think of moving `halo_model` inside `fiducial` together with `cosmo`, `gc_bias`, etc.? My argument is that the halo model is not a property of the covariance, but of the modelling of the power spectrum

xcell/cls/cl.py
```diff
@@ -251,17 +294,23 @@ def compute_tracer_ccl(self, tr, mapper):
         dtype = mapper.get_dtype()
         tracer = self.data.data['tracers'][tr]
         fiducial = self.data.data['cov']['fiducial']
+        ccl_pr = self.hm_par['prof_NFW']
+        ccl_pr_2pt = None
+        with_hm = tracer.get('use_halo_model', False)
+        normed_profile = True
```
carlosggarcia, line 300: Is `normed_profile` always True? If so, why is it interesting to return it? If not, I'd add it as an option to be read from the file.

xcell/cls/cl.py
```diff
@@ -251,17 +294,23 @@ def compute_tracer_ccl(self, tr, mapper):
         dtype = mapper.get_dtype()
         tracer = self.data.data['tracers'][tr]
         fiducial = self.data.data['cov']['fiducial']
+        ccl_pr = self.hm_par['prof_NFW']
+        ccl_pr_2pt = None
```
carlosggarcia, line 298: Is `ccl_pr_2pt` always None? If so, why is it interesting to return it? If not, I'd add it as an option to be read from the file.

xcell/cls/cl.py
```diff
@@ -274,15 +323,58 @@ def compute_tracer_ccl(self, tr, mapper):
...
+                'ccl_tr': ccl_tr,
+                'ccl_pr': ccl_pr,
+                'ccl_pr_2pt': ccl_pr_2pt,
+                'with_hm': with_hm,
+                'normed': normed_profile}
+
+    def _get_ccl_pk(self, ccl_tr1, ccl_tr2):
+        if ccl_tr1['with_hm'] or ccl_tr2['with_hm']:
+            if ccl_tr1['name'] == ccl_tr2['name']:
+                pr2pt = ccl_tr1['ccl_pr_2pt']
```
carlosggarcia, line 345: This would be `None`, at this moment (see L. 298)

xcell/cls/cl.py
```diff
@@ -274,15 +323,58 @@ def compute_tracer_ccl(self, tr, mapper):
...
+            else:
+                pr2pt = None
+            k_s = np.geomspace(1E-4, 1E2, 512)
```
carlosggarcia, line 348: Why geometric?

xcell/cls/cl.py
```diff
@@ -293,7 +385,10 @@ def get_cl_file(self):
         ell = np.arange(3 * nside)
         if not os.path.isfile(fname):
             ccl_tr1, ccl_tr2 = self.get_tracers_ccl()
-            cl = ccl.angular_cl(self.cosmo, ccl_tr1, ccl_tr2, ell)
+            cl = self._get_ccl_cl(ccl_tr1, ccl_tr2, ell)
+            b1 = self.data.data['tracers'][self.tr1].get('bias', 1.)
+            b2 = self.data.data['tracers'][self.tr2].get('bias', 1.)
+            cl *= b1*b2
```
carlosggarcia, line 391: I like this. Very clever.

xcell/tests/test_cls.py
```diff
@@ -382,6 +382,37 @@ def test_cov_spin0(perm):
...
+                                 'halo_bias': hb,
+                                 'concentration': cM}
+    clf = ClFid(data, 'Dummy__0', 'Dummy__1')
+    assert np.fabs(clf.hm_par['mass_def'].get_Delta(clf.cosmo, 1.)
+                   - 200) < 1E-3
+    assert clf.hm_par['mass_def'].rho_type == 'critical'
+    assert clf.hm_par['mass_func'].name == mf
+    assert clf.hm_par['halo_bias'].name == hb
+    assert clf.hm_par['cM'].name == cM
+
```
carlosggarcia, line 414: Add another `shutil.rmtree(tmpdir1)`?

xcell/tests/test_cls.py
```diff
@@ -382,6 +382,37 @@ def test_cov_spin0(perm):
             assert np.all(r[i::ncls1][1:] < 1E-5)
 
 
+def test_clfid_halomod_settings():
+    data = get_config()
+
+    # Empty halo model parameters (default values)
+    clf = ClFid(data, 'Dummy__0', 'Dummy__1')
+    assert np.fabs(clf.hm_par['mass_def'].get_Delta(clf.cosmo, 1.)
```
carlosggarcia, line 390: What does `get_Delta` do?

carlosggarcia (changes requested):
All good. Minor comments/questions.

xcell/cls/cl.py
damonge: Good idea! Done

xcell/cls/cl.py
damonge: It's not always True. Now that we have tSZ, the tSZ profile is not normed (see new code)

xcell/cls/cl.py
damonge: No, check out the HOD case

xcell/cls/cl.py
damonge: nope!

xcell/cls/cl.py
damonge: logarithmic sampling is a good way of capturing the features of the power spectrum

xcell/tests/test_cls.py
damonge: yep, done

xcell/tests/test_cls.py
damonge: returns the spherical overdensity associated with this mass definition (200 in this case)

damonge:
Thanks a lot. Back to you @carlosggarcia

xcell/cls/cl.py
carlosggarcia: Oh, I see! Missed that line. Sorry!

xcell/cls/cl.py
carlosggarcia: Now I see!

xcell/cls/cl.py
carlosggarcia: Oooh! I thought geometric was something else. I just saw the numpy [documentation](https://numpy.org/doc/stable/reference/generated/numpy.geomspace.html) and this is the same thing as logarithmic: 

```
Return numbers spaced evenly on a log scale (a geometric progression).

This is similar to logspace, but with endpoints specified directly. Each output sample is a constant multiple of the previous.
```

xcell/cls/cl.py
```diff
@@ -293,7 +394,10 @@ def get_cl_file(self):
         ell = np.arange(3 * nside)
         if not os.path.isfile(fname):
             ccl_tr1, ccl_tr2 = self.get_tracers_ccl()
-            cl = ccl.angular_cl(self.cosmo, ccl_tr1, ccl_tr2, ell)
+            cl = self._get_ccl_cl(ccl_tr1, ccl_tr2, ell)
+            b1 = self.data.data['tracers'][self.tr1].get('bias', 1.)
```
carlosggarcia, line 398: Should we add here the `if self.data['cov']['fiducial']['gc_bias']` or do you think we should always add the galaxy bias (which I guess 99% we will want that)?

xcell/tests/data/desy1_ebossqso_p18cmbk.yml
```diff
@@ -190,7 +190,6 @@ cov:
 
 
     #### Biases to apply to the tracers ####
-    gc_bias: True
```
carlosggarcia, line 193: See my comment on `cls.py` L.398

carlosggarcia (changes requested):
Almost there. Few more comments

xcell/cls/cl.py
carlosggarcia: I convinced myself that the `if` is not necessary. If you don't want to account for the galaxy bias, just remove it from the tracers or set it to 1.

carlosggarcia (approved):
LGTM

PR #138, opened 2021-08-10
ci: fix cache

carlosggarcia:
The cache sometimes causes errors. Let's fix it!


carlosggarcia:
It seems that it works (I run the action twice to make sure it used the cache). I'll merge it and cross fingers.

PR #141, opened 2021-08-12
Cells from data for covariance

### LINKED ISSUE #135: Covariance from data Cls
Sometimes it'd be useful to pass the data power spectra as the theory ones to compute covariances (especially when the noise properties are weird)

damonge:
Closes #135

### DIFF
--- a/xcell/cls/cov.py
+++ b/xcell/cls/cov.py
@@ -69,11 +69,14 @@ class Cov():
                     # Try to compute the fiducial Cl
                     try:
                         cl = ClFid(data, *trs)
-                    except NotImplementedError:
-                        # If that fails (e.g. unknown data type)
-                        # this will be computed from the data.
-                        cl = cl_dic[trs]
-                        self.tmat[trs]['clcov_from_data'] = True
+                    except NotImplementedError as e:
+                        if self.data.data['cov'].get('data_fallback', False):
+                            # If that fails (e.g. unknown data type)
+                            # this will be computed from the data.
+                            cl = cl_dic[trs]
+                            self.tmat[trs]['clcov_from_data'] = True
+                        else:
+                            raise NotImplementedError(e)
                 clfid_dic[trs] = cl
 
         return cl_dic, clfid_dic
--- a/xcell/cls/cov.py
+++ b/xcell/cls/cov.py
@@ -69,7 +69,7 @@ class Cov():
                     # Try to compute the fiducial Cl
                     try:
                         cl = ClFid(data, *trs)
-                    except ValueError:
+                    except NotImplementedError:
                         # If that fails (e.g. unknown data type)
                         # this will be computed from the data.
                         cl = cl_dic[trs]
--- a/xcell/cls/cov.py
+++ b/xcell/cls/cov.py
@@ -76,7 +76,7 @@ class Cov():
                             cl = cl_dic[trs]
                             self.tmat[trs]['clcov_from_data'] = True
                         else:
-                            raise NotImplementedError(e)
+                            raise e
                 clfid_dic[trs] = cl
 
         return cl_dic, clfid_dic
--- a/xcell/cls/cov.py
+++ b/xcell/cls/cov.py
@@ -9,6 +9,7 @@ import pymaster as nmt
 class Cov():
     def __init__(self, data, trA1, trA2, trB1, trB2):
         self.data = Data(data=data)
+        self.tmat = self.data.get_tracer_matrix()
         self.outdir = self.get_outdir()
         os.makedirs(self.outdir, exist_ok=True)
         self.trA1 = trA1
@@ -61,7 +62,19 @@ class Cov():
         clfid_dic = {}
         for trs in trs_comb:
             if trs not in clfid_dic.keys():
-                clfid_dic[trs] = ClFid(data, *trs)
+                # Cl computed from data if needed
+                if self.tmat[trs]['clcov_from_data']:
+                    cl = cl_dic[trs]
+                else:
+                    # Try to compute the fiducial Cl
+                    try:
+                        cl = ClFid(data, *trs)
+                    except ValueError:
+                        # If that fails (e.g. unknown data type)
+                        # this will be computed from the data.
+                        cl = cl_dic[trs]
+                        self.tmat[trs]['clcov_from_data'] = True
+                clfid_dic[trs] = cl
 
         return cl_dic, clfid_dic
 
@@ -100,8 +113,11 @@ class Cov():
         if not mean_mamb:
             cl_cp = np.zeros_like(nl_cp)
         else:
-            w = clab.get_workspace()
-            cl_cp = w.couple_cell(clab_fid.get_ell_cl()[1])
+            if isinstance(clab_fid, ClFid):  # Compute from theory
+                w = clab.get_workspace()
+                cl_cp = w.couple_cell(clab_fid.get_ell_cl()[1])
+            else:  # Compute from data
+                cl_cp = clab_fid.get_ell_cl_cp()[1]
             cl_cp = (cl_cp + nl_cp) / mean_mamb
 
         return cl_cp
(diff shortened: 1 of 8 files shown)


damonge:
Now tested

xcell/mappers/mapper_dummy.py
```diff
@@ -68,7 +68,7 @@ def __init__(self, config):
 
     def _check_dtype(self):
         dtypes = ['galaxy_density', 'galaxy_shear',
-                  'cmb_convergence', 'cmb_tSZ']
+                  'cmb_convergence', 'cmb_tSZ', 'generic']
         if self.dtype not in dtypes:
             raise ValueError(f'Tracer type {self.dtype} not implemented.')
```
carlosggarcia, line 73: Since you will be using this in `cov.py`, I would update it to NotImplementedError.

xcell/tests/test_cls.py
```diff
@@ -210,8 +214,8 @@ def test_get_covariance():
     chi2 = dCl.dot(icov).dot(dCl)
     chi2_m = dCl.dot(icov_m).dot(dCl)
 
-    assert np.fabs(chi2/chi2_m) - 1 < 0.01
     shutil.rmtree(tmpdir1)
+    assert np.fabs(chi2/chi2_m-1) - 1 < 0.03
```
carlosggarcia, line 218: Why do you need to increase the relative error?

xcell/tests/test_cls.py
```diff
@@ -294,6 +298,13 @@ def test_symmetric():
     shutil.rmtree(tmpdir1)
 
 
+def test_unsupported_quantity():
+    data = get_config(dtype0='generic')
+    with pytest.raises(ValueError):
```
carlosggarcia, line 303: If you agree with me on using `NotImplementedError`, update this

xcell/tests/test_cls.py
```diff
@@ -294,6 +298,13 @@ def test_symmetric():
     shutil.rmtree(tmpdir1)
 
 
+def test_unsupported_quantity():
+    data = get_config(dtype0='generic')
```
carlosggarcia, line 302: `generic` is supported now. Should you use a different one (like `MyVeryUnsupportedTracerType`)?

xcell/tests/test_data.py
```diff
@@ -48,6 +48,72 @@ def remove_outdir(config):
...
+        assert m[(t1, t2)]['clcov_from_data']
+    remove_yml_file(c)
+
+    # Group cls from data (all)
+    c = get_config_dict()
+    c['cov']['cls_from_data'] = {'DESgc-DESgc': {'compute': 'all'}}
+    data = Data(data=c)
+    m = data.get_tracer_matrix()
+    for t1, t2 in get_tracer_pair_iterator(data):
+        if t1[:5] == 'DESgc' and t2[:5] == 'DESgc':
```
carlosggarcia, line 82: You can use `get_tracer_bare_name` https://github.com/xC-ell/xCell/blob/686c4f473d2f6f52bd5f99ebb98903cef7fbf60d/xcell/cls/data.py#L71
or even `get_tracer_bare_name_pair` https://github.com/xC-ell/xCell/blob/686c4f473d2f6f52bd5f99ebb98903cef7fbf60d/xcell/cls/data.py#L76 and do:

```
bname = get_tracer_bare_name_pair(tr1, tr2)
if bname == 'DESgc-DESgc':
```

carlosggarcia (changes requested):
A few comments and suggestions. Nothing critical.

xcell/cls/cov.py
```diff
@@ -61,7 +62,19 @@ def _load_Cls(self):
...
             if trs not in clfid_dic.keys():
-                clfid_dic[trs] = ClFid(data, *trs)
+                # Cl computed from data if needed
+                if self.tmat[trs]['clcov_from_data']:
+                    cl = cl_dic[trs]
+                else:
+                    # Try to compute the fiducial Cl
+                    try:
+                        cl = ClFid(data, *trs)
+                    except ValueError:
```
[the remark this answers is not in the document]
damonge: Should we throw a warning then when this happens? And what do you mean by a flag in the config file?
Happy to change valueerror to notimplemented

xcell/cls/data.py
```diff
@@ -73,6 +73,58 @@ def get_tracer_bare_name(self, tr):
...
+                if pname in clcovlist:
+                    # Out of this group, should this cl be comp. from data?
+                    comp = clcovlist[pname]['compute']
+                    if ((comp == 'all') or
+                            ((comp == 'auto') and (tn1 == tn2))):
+                        clcov_from_data = True
+        return {'compute': compute,
+                'clcov_from_data': clcov_from_data}
+
+    def get_tracer_matrix(self):
```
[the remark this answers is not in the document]
damonge: I thought, since for a general covariance you already need to know about 6 tracer pairs, we might as well just get the full matrix, since it doesn't take much time to build it.

xcell/tests/test_cls.py
damonge: Because now we're comparing with the covariance calculated from the estimated C_ell, which is a bit noisy. I can use two different tolerances for the two modes (note also that the previous test was wrong!)

xcell/tests/test_cls.py
damonge: `generic` is not supported for `ClFid`. And if I changed it to anything else, an error will be thrown because `MapperDummy` does not support this type (and I want to check the `ClFid` error instead)

xcell/cls/data.py
```diff
@@ -73,6 +73,58 @@ def get_tracer_bare_name(self, tr):
...
+    def _get_pair_reqs(self, tn1, tn2):
+        tb1 = self.get_tracer_bare_name(tn1)
+        tb2 = self.get_tracer_bare_name(tn2)
+        pname = f'{tb1}-{tb2}'
+        pname_inv = f'{tb1}-{tb2}'
+        compute = False
+        clcov_from_data = False
+
+        # Find if we want to compute this C_ell
+        if pname_inv in self.data['cls']:
```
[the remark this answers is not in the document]
damonge: OK, I tried doing this but I think it's better as it is. Note that `pname` is e.g. `'DESwl-DESwl'`, whereas the elements of `cls_to_compute` are e.g. `('DESwl__0', 'DESwl__0')`, so I'd need to search for both types of pair names, which adds a lot of extra code.
Anyway, I've tried to solve #143 while looking into this.

xcell/cls/data.py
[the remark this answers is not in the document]
damonge: And now we're saving it as an attribute anyway.

xcell/cls/data.py
```diff
@@ -73,6 +73,58 @@ def get_tracer_bare_name(self, tr):
...
+                if pname_inv in clcovlist:
+                    pname = pname_inv
+                if pname in clcovlist:
+                    # Out of this group, should this cl be comp. from data?
+                    comp = clcovlist[pname]['compute']
+                    if ((comp == 'all') or
+                            ((comp == 'auto') and (tn1 == tn2))):
+                        clcov_from_data = True
+        return {'compute': compute,
+                'clcov_from_data': clcov_from_data}
```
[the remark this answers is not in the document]
damonge: Added `inv`

xcell/mappers/mapper_dummy.py
damonge: Done

xcell/tests/test_cls.py
damonge: Done

xcell/tests/test_data.py
damonge: Done!

damonge:
Back to you @carlosggarcia . Thanks!

xcell/cls/cov.py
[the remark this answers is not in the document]
carlosggarcia: I mean checking the following: `if data.data['cov']['fallback'] == 'data'` or using `True/False` if we don't expect to have more fallback options.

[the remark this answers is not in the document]
damonge: OK, sounds good. Let's have a boolean (`data['cov']['data_fallback'] = True`), which will be `False` by default. Sound good?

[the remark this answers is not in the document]
carlosggarcia: Yep. Thanks

xcell/cls/data.py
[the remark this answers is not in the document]
carlosggarcia: You're right. Sorry for the inconvenience!

xcell/cls/data.py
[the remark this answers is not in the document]
carlosggarcia: Nice, thanks!

xcell/cls/data.py
```diff
@@ -24,6 +24,9 @@ def __init__(self, data_path='', data={}, override=False):
 
         os.makedirs(self.data['output'], exist_ok=True)
         self._check_yml_in_outdir(override)
+        self.tr_matrix = None
+        self.cl_tracers = {True: None, False: None}
```
carlosggarcia, line 28: I was wondering what the `True` and `False` were. Now I see they are the bools for the `wsp` option. I think we should change this to something more readable, but since it will require modifying more parts of the code, we can leave it for a later PR

xcell/tests/test_cls.py
carlosggarcia: you're right! I think we can leave it as is now.

xcell/tests/test_cls.py
carlosggarcia: Understood.

xcell/tests/test_data.py
```diff
@@ -48,6 +48,83 @@ def remove_outdir(config):
...
+            assert not m[(t1, t2)]['clcov_from_data']
+    remove_yml_file(c)
+
+    # Group cls from data (auto)
+    c = get_config_dict()
+    c['cov']['cls_from_data'] = {'DESgc-DESgc': {'compute': 'auto'}}
+    data = Data(data=c)
+    m = data.get_tracer_matrix()
+    for t1, t2 in get_tracer_pair_iterator(data):
+        if t1[:5] == 'DESgc' and t2[:5] == 'DESgc' and t1 == t2:
```
carlosggarcia, line 105: You can use `tpair` here as well!

xcell/cls/cov.py
```diff
@@ -61,7 +62,22 @@ def _load_Cls(self):
...
+                    try:
+                        cl = ClFid(data, *trs)
+                    except NotImplementedError as e:
+                        if self.data.data['cov'].get('data_fallback', False):
+                            # If that fails (e.g. unknown data type)
+                            # this will be computed from the data.
+                            cl = cl_dic[trs]
+                            self.tmat[trs]['clcov_from_data'] = True
+                        else:
+                            raise NotImplementedError(e)
```
carlosggarcia, line 79: hmm. Wouldn't it be `raise e`?

carlosggarcia (changes requested):
Minor comments

xcell/cls/cov.py
damonge: yes, done!

xcell/cls/data.py
damonge: Changed to `wsp` and `no_wsp`

xcell/tests/test_data.py
damonge: Done

carlosggarcia (approved):
LGTM

PR #142, opened 2021-08-12
Cell covariance for auto-correlations with custom methods

damonge:


xcell/cls/cl.py
```diff
@@ -48,6 +48,12 @@ def get_ell_cl(self):
             self.get_cl_file()
         return self.ell, self.cl
 
+    def get_n_cls(self):
```
carlosggarcia, line 51: Nice.

Can you replace calls to `get_dof_tracers`  in `to_sacc.py` with this method? https://github.com/xC-ell/xCell/blob/d37b89b7fd5c147d02bde9b4f70dcea937c4a774/xcell/cls/to_sacc.py#L192

xcell/cls/cl.py
```diff
@@ -173,24 +180,35 @@ def get_cl_file(self):
...
-                nl = np.zeros([nmap1 * nmap2, self.b.get_n_bands()])
+                n_cls = self.get_n_cls()
+                nl_cp = np.zeros((n_cls, 3 * self.nside))
+                nl = np.zeros([n_cls, self.b.get_n_bands()])
             # Signal
             if auto and mapper1.custom_auto:
                 cl_cp = mapper1.get_cl_coupled()
                 cl = w.decouple_cell(cl_cp)
+                cls_cov = mapper1.get_cls_covar_coupled()
+                cl_cov_cp = cls_cov['cross']
```
carlosggarcia, line 191: Could you add a comment here about these new terms?

xcell/cls/cl.py
```diff
@@ -173,24 +180,35 @@ def get_cl_file(self):
...
+                cl_cov_cp = nmt.compute_coupled_cell(f1, f2)
+                cl_cov_11_cp = cl_cov_cp
+                cl_cov_12_cp = cl_cov_cp
+                cl_cov_22_cp = cl_cov_cp
+                cl = w.decouple_cell(cl_cov_cp)
+                cl_cp = cl_cov_cp - nl_cp
                 cl -= nl
             np.savez(fname, ell=ell, cl=cl, cl_cp=cl_cp, nl=nl,
-                     nl_cp=nl_cp, wins=wins)
+                     nl_cp=nl_cp, cl_cov_cp=cl_cov_cp,
```
carlosggarcia, line 207: We should comment somewhere that `cl_cov_X` includes the noise contribution

xcell/cls/cl.py
```diff
@@ -173,24 +180,35 @@ def get_cl_file(self):
...
                 # the noise-less power spectrum. No need
                 # to subtract it here.
             else:
-                cl_cp = nmt.compute_coupled_cell(f1, f2)
-                cl = w.decouple_cell(cl_cp)
-                cl_cp -= nl_cp
+                cl_cov_cp = nmt.compute_coupled_cell(f1, f2)
+                cl_cov_11_cp = cl_cov_cp
+                cl_cov_12_cp = cl_cov_cp
+                cl_cov_22_cp = cl_cov_cp
```
carlosggarcia, line 202: Not sure I understand why 11, 12 and 22 are all the same if tr1 and tr2 will be different.

xcell/cls/cov.py
```diff
@@ -112,16 +112,16 @@ def get_covariance_workspace(self):
...
-        nl_cp = clab.get_ell_nl_cp()[1]
         if not mean_mamb:
-            cl_cp = np.zeros_like(nl_cp)
+            cl_cp = np.zeros((clab.get_n_cls(), 3*clab.nside))
         else:
             if isinstance(clab_fid, ClFid):  # Compute from theory
                 w = clab.get_workspace()
                 cl_cp = w.couple_cell(clab_fid.get_ell_cl()[1])
+                cl_cp += clab.get_ell_nl_cp()[1]
             else:  # Compute from data
```
carlosggarcia, line 122: Could you add a comment here reminding me that at this stage, if it hasn't failed, it's because you have requested in the config file to fallback to the data Cell?

xcell/cls/cov.py
```diff
@@ -112,16 +112,16 @@ def get_covariance_workspace(self):
...
-                cl_cp = clab_fid.get_ell_cl_cp()[1]
-            cl_cp = (cl_cp + nl_cp) / mean_mamb
+                cl_cp = clab_fid.get_ell_cl_cp_cov()[1]
```
carlosggarcia, line 123: Does this work? if you're inside this `else` is because `clab_fid` is not a instance of the `ClFid` class

xcell/mappers/mapper_dummy.py
```diff
@@ -189,6 +190,15 @@ def get_cl_coupled(self):
             self.cl_coupled += self.custom_offset
         return self.cl_coupled
 
+    def get_cls_covar_coupled(self):
+        if self.cls_cov is None:
+            clc = self.get_cl_coupled()
+            self.cls_cov = {'cross': clc,
+                            'auto_11': clc,
```
carlosggarcia, line 197: Wouldn't it be better that each of these were different? My point is that if 'auto' is not True in the `ClFid` (L. 192), then the `cl_cov_cp` will be the same. It might make the tests a bit confusing.

xcell/tests/test_cls.py
```diff
@@ -167,6 +167,16 @@ def test_custom_auto():
...
 
+    # Covariance custom cross
+    data = get_config()
+    data['tracers']['Dummy__0']['custom_auto'] = True
+    data['tracers']['Dummy__0']['custom_offset'] = np.pi*1E-5
+    clc3 = Cl(data, 'Dummy__0', 'Dummy__0')
+    l2, cl3 = clc3.get_ell_cl_cp_cov()
+    shutil.rmtree(tmpdir1)
+
+    assert np.allclose(cl1, cl2-np.pi*1E-5, rtol=1E-4, atol=0)
```
carlosggarcia, line 178: `cl3`?

xcell/tests/test_cls.py
```diff
@@ -167,6 +167,16 @@ def test_custom_auto():
 
     assert np.allclose(cl1, cl2-np.pi*1E-5, rtol=1E-4, atol=0)
 
+    # Covariance custom cross
```
carlosggarcia, line 170: Missing test for the new parts of the method and the new output of the cl.npz file.

carlosggarcia (changes requested):
Some questions and changes. I think there are missing tests: for mapper_dummy and for the new additions to ClFid.

xcell/cls/cov.py
damonge: Exacly, in this case `clab_fid` is a `Cl`, which has a `get_ell_cl_cp_cov` method.

xcell/tests/test_cls.py
damonge: Aren't these being tested in https://github.com/xC-ell/xCell/blob/ed32e21154b7e85aa10f169b0eb8c970cf99cf7d/xcell/tests/test_mapper_planck_all.py#L84?
What do you want to test about the `npz` files?

xcell/cls/cl.py
damonge: Well, the input of both functions is not the same, so instead I've streamlined `get_dof_tracers`

xcell/cls/cl.py
damonge: OK, I've added an extensive comment explaining what this PR is supposed to implement.

xcell/cls/cl.py
damonge: Done

xcell/cls/cl.py
damonge: See the extended comment I added in the code and note that these C_ells are only used if both tracers are the same (i.e. this is an auto correlation), and when we're computing the auto-covariance of this auto-correlation. In that case, if no custom method exists, the covariance is proportional to `(C^11*C^22+C^12^2) = 2C^aa^2` (i.e. all C_ells coincide).

xcell/cls/cov.py
damonge: Done

xcell/mappers/mapper_dummy.py
damonge: OK, I've subtracted `custom_offset` from `cross` and `auto_12` so we can test that they are different.

xcell/tests/test_cls.py
damonge: oops! Yes, done

carlosggarcia (changes requested):
Minor comment/question. Needed test for `MapperDummy` new method `get_cls_covar_coupled`.

xcell/tests/test_cls.py
damonge: Done

damonge:
Done @carlosggarcia

carlosggarcia (approved):
LGTM

PR #145, opened 2021-08-13
Theory module

damonge:


damonge:
@carlosggarcia I changed the scope of this PR to make it easier to review the next one (non-Gaussian covariances). Can you take a quick look and tell me what you think? The idea is factoring everything to do with theory prediction into its own separate class that can be shared between `ClFid` and  the future `CovNGFid`

xcell/cls/cl.py
```diff
@@ -261,160 +261,22 @@ def __init__(self, data, tr1, tr2):
 
         self.outdir = self.get_outdir('fiducial')
         os.makedirs(self.outdir, exist_ok=True)
-        self.cosmo = self.get_cosmo_ccl()
-        self.hm_par = self._get_halomodel_params()
+        self.th = Theory(self.data.data)
```
carlosggarcia, line 264: Why don't you make `ClFid` a child also of Theory? (i.e. `class ClFid(ClBase, Theory)`)?

xcell/cls/cl.py
```diff
@@ -261,160 +261,22 @@ def __init__(self, data, tr1, tr2):
...
-                'calculator': hmc,
-                'alpha': (lambda a: alpha),
-                'k_suppress': (lambda a: klow)}
-
-    def get_cosmo_ccl(self):
-        fiducial = self.data.data['cov']['fiducial']
-        cosmo = ccl.Cosmology(**fiducial['cosmo'])
-        return cosmo
-
     def get_tracers_ccl(self):
```
carlosggarcia, line 268: Why don't you move this to `Theory` as well?

carlosggarcia (commented):
I like this structure. I would just make `ClFid` a child of `Theory` as well.

xcell/cls/cl.py
damonge: Fancy! Let me see if I can make that work...

xcell/cls/cl.py
damonge: Because this is specific of power spectra, which involve a pair of tracers. There'll be something like that for covariances too.

xcell/cls/cl.py
damonge: Looking into it I'm not sure this is a good idea. For this to work, `ClBase` and `Theory` would need to have the same `__init__()` syntax, which they don't. I'd rather not force the code to do this just for the sake of multiple inheritance fanciness.

carlosggarcia (changes requested):
Looks good but would it be too much work to add/split the test for `Theory`?

damonge:
I've just moved lines of code around. So our tests are effectively testing the same code. Happy to open an issue entitled "organise tests"

carlosggarcia (approved):

carlosggarcia:
Sounds good

PR #146, opened 2021-08-16
Non-Gaussian halo-model covariances

damonge:


xcell/cls/cov.py
```diff
@@ -429,39 +431,90 @@ def get_covariance(self):
...
 
-        if self.nl_marg:
-            cov += self.get_covariance_nl_marg()
+        if self.m_marg and notnull:
+            cov_mm = self.get_covariance_m_marg()
 
-        if self.m_marg:
-            cov += self.get_covariance_m_marg()
+        if self.do_NG and notnull:
+            fsky = self.data.data['cov'].get('fsky_NG', -1.)
```
carlosggarcia, line 468: Not sure I like using numbers for these cases. I don't think the meaning will be obvious.  I'd prefer to use `None`.

xcell/cls/cov.py
```diff
@@ -429,39 +431,90 @@ def get_covariance(self):
...
+                                                           fsky, kind)
 
-        self.cov = cov
-        np.savez_compressed(fname, cov=cov)
+        self.cov = cov_G + cov_nlm + cov_mm + cov_NG
+        np.savez_compressed(fname, cov=self.cov, cov_G=cov_G, cov_NG=cov_NG,
+                            cov_nl_marg=cov_nlm, cov_m_marg=cov_mm)
         self.recompute_cov = False
-        return cov
+        return self.cov
```
carlosggarcia, line 483: Good catch! Thanks!

xcell/cls/cov.py
```diff
@@ -429,39 +431,90 @@ def get_covariance(self):
...
+
+    def get_covariance_ng_halomodel(self, s_a1, s_a2, s_b1, s_b2,
+                                    fsky, kind='1h'):
+        ellA = self.clA1A2.b.get_effective_ells()
+        ellB = self.clB1B2.b.get_effective_ells()
+        nclsa = np.max([1, s_a1 + s_a2])
+        nclsb = np.max([1, s_b1 + s_b2])
+        cov = np.zeros([len(ellA), nclsa, len(ellB), nclsb])
```
carlosggarcia, line 491: If `ellA` is a numpy array `ellA.size` is faster, i think

xcell/cls/cov.py
```diff
@@ -429,39 +431,90 @@ def get_covariance(self):
...
+        th = Theory(self.data.data)
+        mpA1, mpA2 = self.clA1A2.get_mappers()
+        mpB1, mpB2 = self.clB1B2.get_mappers()
+        trlist = self.data.data['tracers']
+        ccl_trA1 = th.compute_tracer_ccl(self.trA1,
+                                         trlist[self.trA1],
+                                         mpA1)
+        bA1 = trlist[self.trA1].get('bias', 1.)
```
carlosggarcia, line 499: We should unify the way we get the bias, to avoid doing `.get('bias', 1)` in every place.

xcell/cls/theory.py
```diff
@@ -119,10 +119,49 @@ def compute_tracer_ccl(self, name, tracer, mapper):
                 'with_hm': with_hm,
                 'normed': normed_profile}
 
+    def get_ccl_tkka(self, ccl_trA1, ccl_trA2, ccl_trB1, ccl_trB2,
```
carlosggarcia, line 122: Can you add a comment saying what is tkka?

xcell/cls/theory.py
```diff
@@ -119,10 +119,49 @@ def compute_tracer_ccl(self, name, tracer, mapper):
...
+            pr2ptA = None
+        pB1 = ccl_trB1['ccl_pr']
+        pB2 = ccl_trB2['ccl_pr']
+        if ccl_trB1['name'] == ccl_trB2['name']:
+            pr2ptB = ccl_trB1['ccl_pr_2pt']
+            pB2 = pB1
+        else:
+            pr2ptB = None
+
+        k_s = np.geomspace(1E-4, 1E2, 512)
```
carlosggarcia, line 143: Is it ok having it hardcoded?

xcell/cls/theory.py
```diff
@@ -119,10 +119,49 @@ def compute_tracer_ccl(self, name, tracer, mapper):
...
+                                        a_arr=a_s, lk_arr=lk_s)
+        return tkk
+
     def get_ccl_pk(self, ccl_tr1, ccl_tr2):
         if ccl_tr1['with_hm'] or ccl_tr2['with_hm']:
+            p1 = ccl_tr1['ccl_pr']
+            p2 = ccl_tr2['ccl_pr']
             if ccl_tr1['name'] == ccl_tr2['name']:
                 pr2pt = ccl_tr1['ccl_pr_2pt']
+                p2 = p1
```
carlosggarcia, line 164: hmmm... if they're the same tracer, it guess it is ok that `p2` is exactly the same dictionary as `p1` (it points to the same memory location). However, could there be any case when `p2` might be later modified and, therefore, be different to `p1`? If so, use `p1.copy()`

carlosggarcia (changes requested):
Minor comments

xcell/cls/theory.py
damonge: Note that `p1` and `p2` are not dictionaries, but `ccl.halos.HaloProfile` objects, and if both are not **exactly** the same, ccl will throw an error if you attempt to use a `Profile2ptHOD` object for the auto-correlation. So yes, they need to be exactly the same profile object (at the memory level), and `.copy()` won't work.

xcell/cls/theory.py
damonge: For all supported tracers this is accurate enough while being manageable (everything computes in under 1s). Note we're already doing the same thing for the halo model power spectra.

We can make it fully tunable in the future if we need to.

xcell/cls/cov.py
damonge: What do you propose?

carlosggarcia: Maybe a method in `Data`? I don't like it too much, but since it is related to reading the configuration of the tracers, it might not be too odd.

xcell/cls/theory.py
damonge: Done

xcell/cls/cov.py
damonge: Done

xcell/cls/cov.py
damonge: Done

xcell/cls/cov.py
damonge: Done

carlosggarcia: Thanks

xcell/cls/cov.py
carlosggarcia: Great! Thanks!

carlosggarcia (approved):
LGTM

PR #147, opened 2021-09-02
LOW-Z and kappa fixes

damonge:
A few fixes and upgrades after testing the 2MPZ, WIxSC and Planck18 kappa mappers in practice


damonge:
Here's a summary of what I've done here:
- Added a concentration-mass relation that is not currently part of CCL but which we need for the tSZ mapper.
- Refactored the theory calculation slightly so that the cosmology does not get regenerated if it's not needed (accelerates `to_sacc` significantly).
- Updated `to_sacc.py` to include tSZ and generic maps.
- Moved 2MPZ and WIxSC to fitsio, since astropy didn't really work well with those files after the latest version.
- Allowed 2MPZ and WIxSC to be used in galactic coordinates
- Adjusted redshift ranges for both mappers.

damonge:
OK, @carlosggarcia . This is ready for review when you get a chance. Have a look at my previous comment for a guide for what's changed.

damonge:
@carlosggarcia can we merge this?

input/yxgxk.yml
```diff
@@ -0,0 +1,234 @@
+tracers:
```
carlosggarcia, line 1: Isn't there a more obscure file name? What `yxgxk` stands for? `ymilca` cross `g`? cross `kappa`?

xcell/cls/theory.py
```diff
@@ -2,21 +2,54 @@
...
+            return True
+        return False
+
+    def _setup(self):
+        self.A = 3.67
+        self.B = -0.0903
+        self.C = -0.51
+
+    def _concentration(self, cosmo, M, a):
+        M_pivot_inv = cosmo.cosmo.params.h * 5E-13
```
carlosggarcia, line 33: Although easy to find, might you say that this is Eq. 4 in (arXiv:0804.2486)?

xcell/cls/theory.py
```diff
@@ -2,21 +2,54 @@
...
+        return self.A * (M * M_pivot_inv)**self.B * a**(-self.C)
+
+
 class Theory():
     def __init__(self, data):
         self.config = data['cov']['fiducial']
         self.cosmo = None
-        self.get_cosmo_ccl()
```
carlosggarcia, line 9: Good catch

xcell/cls/theory.py
```diff
@@ -2,21 +2,54 @@
...
             self.cosmo = ccl.Cosmology(**(self.config['cosmo']))
         return self.cosmo
 
     def get_halomodel_params(self):
-        if self.cosmo is None:
-            self.get_cosmo_ccl()
+        if self.hm_par is not None:
+            return self.hm_par
+
+        self.get_cosmo_ccl()
```
carlosggarcia, line 52: Maybe it's clearer if you substitute `self.cosmo` in this function for `cosmo` and `cosmo = self.get_cosmo_ccl()`. Same in the other functions. We might forget in a few weeks that `get_cosmo_ccl` is also setting `self.cosmo`.

xcell/cls/theory.py
```diff
@@ -57,19 +91,23 @@ def get_halomodel_params(self):
...
+                       'halo_bias': hb,
+                       'cM': cm,
+                       'prof_NFW': pNFW,
+                       'prof_2pt': p2pt,
+                       'calculator': hmc,
+                       'alpha': (lambda a: alpha),
+                       'k_suppress': (lambda a: klow)}
 
     def compute_tracer_ccl(self, name, tracer, mapper):
+        self.get_cosmo_ccl()
```
carlosggarcia, line 105: Same comment as above

xcell/cls/theory.py
carlosggarcia: It can even make sense to remove `self.cosmo` and use `self._cosmo` to store the cosmology and only access to this by `get_cosmo_ccl`

xcell/cls/theory.py
```diff
@@ -126,21 +164,22 @@ def get_ccl_tkka(self, ccl_trA1, ccl_trA2, ccl_trB1, ccl_trB2,
             raise NotImplementedError(f"Non-Gaussian term {kind} "
                                       "not supported.")
 
+        self.get_cosmo_ccl()
```
carlosggarcia, line 167: Same as above

xcell/cls/theory.py
```diff
@@ -157,6 +196,9 @@ def get_ccl_tkka(self, ccl_trA1, ccl_trA2, ccl_trB1, ccl_trB2,
         return tkk
 
     def get_ccl_pk(self, ccl_tr1, ccl_tr2):
+        self.get_cosmo_ccl()
+        self.get_halomodel_params()
```
carlosggarcia, line 200: Same as with `get_cosmo_ccl`. It might make sense to use `hm_par = self.get_halomodel_params()` and store it internally in `self._hm_par`

xcell/cls/theory.py
```diff
@@ -184,6 +226,8 @@ def get_ccl_pk(self, ccl_tr1, ccl_tr2):
         return pk
 
     def get_ccl_cl(self, ccl_tr1, ccl_tr2, ell):
+        self.get_cosmo_ccl()
```
carlosggarcia, line 229: Same as above

xcell/cls/theory.py
```diff
@@ -193,6 +237,8 @@ def get_ccl_cl(self, ccl_tr1, ccl_tr2, ell):
     def get_ccl_cl_covNG(self, ccl_trA1, ccl_trA2, ellA,
                          ccl_trB1, ccl_trB2, ellB, fsky,
                          kind='1h'):
+        self.get_cosmo_ccl()
```
carlosggarcia, line 240: Same as above

xcell/cls/theory.py
```diff
@@ -203,4 +249,4 @@ def get_ccl_cl_covNG(self, ccl_trA1, ccl_trA2, ellA,
                                       tkka=tkk, fsky=fsky,
                                       cltracer3=ccl_trB1['ccl_tr'],
                                       cltracer4=ccl_trB2['ccl_tr'],
-                                      ell2=ellB)
+                                      ell2=ellB, integration_method='spline')
```
carlosggarcia, line 252: Should we make `integration_method` an input option?

xcell/mappers/mapper_2MPZ.py
```diff
@@ -20,6 +19,7 @@ def __init__(self, config):
         """
         self._get_defaults(config)
         self.z_edges = config.get('z_edges', [0, 0.5])
+        self._get_coords(config)
```
carlosggarcia, line 22: Since you're not getting anything, maybe call it `load_coords` or, even better, make `_get_coords` return `ra_name, dec_name` and in this line you do `self.ra_name, self.dec_name = self._get_coords()`

xcell/mappers/mapper_WIxSC.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, config):
         """
         self._get_defaults(config)
         self.z_edges = config.get('z_edges', [0, 0.5])
+        self._get_coords(config)
```
carlosggarcia, line 27: Same as in `mapper_2MPZ`

xcell/mappers/mapper_WIxSC.py
```diff
@@ -123,8 +141,8 @@ def get_nz(self, dz=0, return_jk_error=False):
                 z, nz, nz_jk = get_DIR_Nz(c_s, c_p,
                                           ['W1c', 'W2c', 'Bcc', 'Rcc'],
                                           zflag='Zspec',
-                                          zrange=[0, 1.],
-                                          nz=100,
+                                          zrange=[0, 0.6],
+                                          nz=150,
```
carlosggarcia, line 145: Why 150 instead of 100?

carlosggarcia (changes requested):
Minor comments. Some suggestions

xcell/mappers/mapper_WIxSC.py
damonge: Just keeping the redshift spacing the same as 2MPZ (100 points from 0 to 0.4)

input/yxgxk.yml
damonge: Yes. It's the name Nick gave to his project https://github.com/nikfilippas/yxgxk. It'll be easier to identify that way, I think.

xcell/cls/theory.py
damonge: Sure, done (but note that the paper was cited in the class docstring)

xcell/cls/theory.py
damonge: There are lots of other default options in these CCL functions we're not letting users modify. This is just because it ensures that the calculation of the covNG won't fail because of a gsl integration error. In 99.9% of the cases that doesn't happen, but because we need to compute so many covariance matrices, I thought it best to use spline integration to avoid the odd case when it happens.

carlosggarcia: Ok. Makes sense. Thanks!

xcell/mappers/mapper_WIxSC.py
damonge: Done

xcell/mappers/mapper_2MPZ.py
damonge: done

xcell/cls/theory.py
damonge: Done

xcell/cls/theory.py
damonge: Done

xcell/cls/theory.py
damonge: Done

xcell/cls/theory.py
damonge: Done

xcell/cls/theory.py
damonge: yep, done

xcell/cls/theory.py
damonge: Done

damonge:
Thanks @carlosggarcia . Back to you

carlosggarcia (approved):
LGTM, thanks!

PR #158, opened 2021-10-07
xcell installable

carlosggarcia:
Made xcell a library and pip installable

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
         #    # Based on: https://medium.com/ai2-blog/python-caching-in-github-actions-e9452698e98d
         #    path: ${{ env.pythonLocation }}
         #    # Use hashFiles to regenerate cache when adding new dependencies
-        #    key: ${{ runner.os }}-cache-${{ hashFiles('.github/workflows/requirements.txt') }}
+        #    key: ${{ runner.os }}-cache-${{ hashFiles('requirements.txt') }}
 
         - name: Install GSL, FFTW (linux)
           run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -41,12 +41,14 @@ jobs:
             python -m pip install -U pip
             # Install wheel to avoid recompiling pymaster, pyccl, camb, etc.
             pip install -U wheel
-            # pip install -U -r .github/workflows/requirements.txt
+            # Install other CI dependencies
+            pip install -U pytest pytest-cov coveralls
```
damonge, line 45: why did you have to take this out of `requirements.txt`?

setup.py
```diff
@@ -0,0 +1,28 @@
...
+    long_description = f.read()
+
+with open("requirements.txt", 'r') as f:
+    requirements = [i.strip() for i in f.readlines()]
+
+setup(
+    name="xCell",
+    version="dev",
+    description=("Pipeline to compute the Nx2pt angular power spectra and their covariances."),
+    license="GPL3",
```
damonge, line 15: Did we add this license file? Also, I think we shouldn't use GPL3 because it's more restrictive than BSD3 (e.g. https://github.com/LSSTDESC/NaMaster/blob/master/LICENSE)

carlosggarcia: Yep. At the beginning of time. I will not make much deal about it but I would keep it GPL because it's more restrictive: I think it's a good thing forcing others not to enclose their code if they use public libraries.

damonge: It spreads like a virus. There may be reasons why people want to use other free software licenses (e.g. because their code has applications in the industry and otherwise it won't be used). As an example, the use of GPL3 by healpix makes it a nightmare for us to use it as a dependency in namaster.

I am fine with GPL, by the way, just not GPL3 (GPL2 is fine)

damonge: That constraint is only there if you use code in your own code, not if you use it as an external dependency (i.e. if we had copy-pasted parts of healpy into our code, or if we were compiling healpy as part of our code, or if we were writing a modified version of healpy)

damonge: if you like GPL let's do GPL2

setup.py
```diff
@@ -0,0 +1,28 @@
...
+with open("README.md", 'r') as f:
+    long_description = f.read()
+
+with open("requirements.txt", 'r') as f:
+    requirements = [i.strip() for i in f.readlines()]
+
+setup(
+    name="xCell",
+    version="dev",
+    description=("Pipeline to compute the Nx2pt angular power spectra and their covariances."),
```
damonge, line 14: can you remove "the" here? Merge after that!

carlosggarcia: Done. Merging! Thanks!

PR #159, opened 2021-10-07
Ignore existing yml when called inside another class

### LINKED ISSUE #112: persistent yml files
Some of the consequences of reading parameters from a yaml file if it exists (rather than taking them from the input dictionary) can be a bit confusing. For instance, if you create a `Covariance` object, while it is initialized, it will create `Cl` objects, which will then throw a warning, because they will read parameters from the yaml file created by the `Covariance` objects rather than the dictionary. Is this intended? Do we want to add a flag that says `ignore_existing_yaml` or something like that for these cases?

carlosggarcia:
The idea of checking if a yml file exists already makes sense at the top Class. Once checked there, there is no need to reread the file. One can pass directly the dictionary. That reduces I/O and confusing warnings.

Closes #112


xcell/cls/data.py
```diff
@@ -23,31 +24,48 @@ def __init__(self, data_path='', data={}, override=False):
...
         fname = os.path.join(outdir, '*.yml')
         files = glob(fname)
-        if (len(files) == 1) and (not override):
+        # Start if-else with sanity checks
+        if override and ignore_existing_yml:
+            raise ValueError('Only one of override or ignore_existing_yml can '
+                             + 'be set.')
+        elif len(files) > 1:
+            raise ValueError(f'More than 1 YML file in outdir: {outdir}.')
+        elif override:
```
damonge, line 45: I'm not sure what the difference between `override` and `ignore_existing_yml` are. I mean I can see they do different things, but I don't understand why you need both of them

damonge: It might also be nice (potentially as part of this PR) to make the code crash if there's already a yaml file that is not the same as the input one. This is what cobaya does (although comparing yaml files may not be trivial). We can discuss offline.

damonge (approved):

PR #160, opened 2021-10-08
healpy section rename

### LINKED ISSUE #109: healpy section - change name
The name of the `healpy` section doesn't make much sense to me, since we don't really use healpy for SHTs. I'd call it `sht` or something like that.

carlosggarcia:
healpy section rename + nside unification under that section.

Closes #109

### DIFF
--- a/xcell/cls/data.py
+++ b/xcell/cls/data.py
@@ -250,7 +250,7 @@ class Data():
     def get_mapper(self, tr):
         config = self.data['tracers'][tr]
         if 'nside' not in config:
-            config['nside'] = self.data['healpy']['nside']
+            config['nside'] = self.data['sphere']['nside']
         mapper_class = config['mapper_class']
         return mapper_from_name(mapper_class)(config)
 
--- a/xcell/cls/data.py
+++ b/xcell/cls/data.py
@@ -249,6 +249,8 @@ class Data():
 
     def get_mapper(self, tr):
         config = self.data['tracers'][tr]
+        if 'nside' not in config:
+            config['nside'] = self.data['healpy']['nside']
         mapper_class = config['mapper_class']
         return mapper_from_name(mapper_class)(config)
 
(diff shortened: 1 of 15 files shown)


damonge:
How about calling it `sphere`?

damonge:
Ok, I understand the renaming, but not the role of `nside`. Right now `nside` could be set ether in `sphere` or in each mapper, and there's no checks that they are compatible, right?

damonge:
OK, but we still need to provide nside as part of the config of each mapper, right? Why not make nside an extra argument of the mapper constructors?

damonge:
@carlosggarcia what I meant here is that we should only provide nside once in the config file. Right now we need to provide it for each mapper and for the healpy section, correct?

xcell/cls/data.py
```diff
@@ -249,6 +249,12 @@ def check_toeplitz(self, dtype):
 
     def get_mapper(self, tr):
         config = self.data['tracers'][tr]
+        nside = self.data['sphere']['nside']
+        if 'nside' not in config:
+            config['nside'] = nside
```
damonge, line 254: I.e. I understand now what you're doing here is adding nside to the mapper's config if it's not there. But this only works if you're gonna get cls. If you just want to instantiate a specific mapper, you need to create a config dictionary that includes `nside`, right?

What I propose is that we change the the constructor of all mappers so that instead of taking only a `config` dictionary, they take `__init__(config, nside)`.

damonge: But maybe I'm being pedantic, we can just drop it...

carlosggarcia: But if your going to instantiate a specific mapper, you will have to create the config dictionary anyway, won't you? What's the advantage you see on making it a separate argument of the constructor?

damonge (approved):
You're right. Let's just merge.

Issue #73, opened 2021-04-22
mappers: Add documentation


carlosggarcia:
We should add information about where the data is taken from, what papers we follow, etc.

carlosggarcia:
DECALS: https://arxiv.org/pdf/2010.00466.pdf and https://arxiv.org/pdf/2010.04698.pdf

JaimeRZP:
I think this issue has been addressed?

PR #162, opened 2021-10-17
HSC DR1 shear sample

damonge:
From https://arxiv.org/abs/2010.09717 and https://arxiv.org/abs/1809.09148

### DIFF
--- a/xcell/mappers/mapper_HSC_DR1wl.py
+++ b/xcell/mappers/mapper_HSC_DR1wl.py
@@ -14,7 +14,9 @@ class MapperHSCDR1wl(MapperBase):
         self.z_edges = config.get('z_edges', [0, 0.5])
         self.bn = self.config['bin_name']
         self.w_name = 'ishape_hsm_regauss_derived_shape_weight'
+        self.npix = hp.nside2npix(self.nside)
 
+        self.nl_coupled = None
         self.dndz = None
         self.cat = None
         self.mask = None
@@ -218,7 +220,7 @@ class MapperHSCDR1wl(MapperBase):
             w2s2 = hp.read_map(fname_lite)
         else:
             cat = self.get_catalog()
-            w2s2 = get_map_from_points(cat, self.side,
+            w2s2 = get_map_from_points(cat, self.nside,
                                        w=(0.5*(cat['e1']**2 + cat['e2']**2) *
                                           cat[self.w_name]**2),
                                        ra_name='ra', dec_name='dec')
--- a/xcell/mappers/mapper_HSC_DR1wl.py
+++ b/xcell/mappers/mapper_HSC_DR1wl.py
@@ -1,0 +1,275 @@
+from .mapper_base import MapperBase
+from .utils import get_map_from_points
+from astropy.table import Table, vstack
+import os
+import numpy as np
+import healpy as hp
+
+
+class MapperHSCDR1wl(MapperBase):
+    def __init__(self, config):
+        self._get_defaults(config)
+        self.path_lite = config.get('path_lite', None)
+        self.icut = config.get('depth_cut', 24.5)
+        self.z_edges = config.get('z_edges', [0, 0.5])
+        self.bn = self.config['bin_name']
+        self.w_name = 'ishape_hsm_regauss_derived_shape_weight'
+
+        self.dndz = None
+        self.cat = None
+        self.mask = None
+        self.signal_map = None
+
+    def _check_lite_exists(self, file_name):
+        if self.path_lite is None:
+            return False, None
+        else:
+            fname_lite = os.path.join(self.path_lite, file_name)
+            return os.path.isfile(fname_lite), fname_lite
+
+    def get_catalog(self):
+        if self.cat is None:
+            file_name = f'HSCDR1wl_{self.bn}.fits'
+            read_lite, fname_lite = self._check_lite_exists(file_name)
+
+            if read_lite:
+                print(f'Loading lite catalog {file_name}')
+                self.cat = Table.read(fname_lite, memmap=True)
+                return self.cat
+
+            cats = []
+            for f in self.config['data_catalogs']:
+                cat = self._clean_raw_catalog(f)
+
+                # Shear cut
+                isn = 'ishape_hsm_regauss'
+                ishape_flags_mask = ~cat[f'{isn}_flags']
+                ishape_sigma_mask = ~np.isnan(cat[f'{isn}_sigma'])
+                ishape_res_mask = cat[f'{isn}_resolution'] >= 0.3
+                ishape_shear_mod_mask = (cat[f'{isn}_e1']**2 +
+                                         cat[f'{isn}_e2']**2) < 2
+                ishape_sigma_mask *= ((cat[f'{isn}_sigma'] >= 0.) *
+                                      (cat[f'{isn}_sigma'] <= 0.4))
+                # Remove masked objects
+                imsk = 'iflags_pixel_bright'
+                star_mask = np.logical_not(cat[f'{imsk}_object_center'])
+                star_mask *= np.logical_not(cat[f'{imsk}_object_any'])
+                fdfc_mask = cat['wl_fulldepth_fullcolor']
+                shearmask = ishape_flags_mask *\
+                    ishape_sigma_mask *\
+                    ishape_res_mask *\
+                    ishape_shear_mod_mask*star_mask *\
+                    fdfc_mask
+                cat = cat[shearmask]
+
+                # Redshift bin cut
+                zs = cat['pz_best_eab']
+                zbinmask = (zs <= self.z_edges[1]) & (zs > self.z_edges[0])
+                cat = cat[zbinmask]
+
+                # Calibrate shear
+                w = cat[self.w_name]
+                mhat = np.average(cat[f'{isn}_derived_shear_bias_m'],
+                                  weights=w)
+                resp = 1. - np.average(cat[f'{isn}_derived_rms_e'] ** 2,
+                                       weights=w)
+                e1 = (cat[f'{isn}_e1']/(2.*resp) -
+                      cat[f'{isn}_derived_shear_bias_c1']) / (1 + mhat)
+                e2 = (cat[f'{isn}_e2']/(2.*resp) -
+                      cat[f'{isn}_derived_shear_bias_c2']) / (1 + mhat)
+                cat['e1'] = e1
+                cat['e2'] = e2
+
+                # Remove unnecessary columns
+                cat.keep_columns(['ra', 'dec', 'e1', 'e2', self.w_name])
+                cats.append(cat)
+            self.cat = vstack(cats)
+            if fname_lite is not None:
+                self.cat.write(fname_lite)
+        return self.cat
+
+    def _clean_raw_catalog(self, fnames):
+        cats = []
+        for fname in fnames:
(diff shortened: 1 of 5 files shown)


xcell/mappers/mapper_HSC_DR1wl.py
```diff
@@ -0,0 +1,278 @@
...
+        for fname in fnames:
+            if not os.path.isfile(fname):
+                raise ValueError(f"File {fname} not found")
+            c = Table.read(fname)
+            sel = np.ones(len(c), dtype=bool)
+            isnull_names = []
+            for key in c.keys():
+                if key.__contains__('isnull'):
+                    if not key.startswith('ishape'):
+                        sel[c[key]] = 0
```
carlosggarcia, line 105: `sel` is of bool type. Shouldn't it be set to `False`? Save for L. 111

xcell/mappers/mapper_HSC_DR1wl.py
```diff
@@ -0,0 +1,278 @@
...
+                    isnull_names.append(key)
+                else:
+                    # Keep photo-zs and shapes even if they're NaNs
+                    if ((not key.startswith("pz_")) and
+                            (not key.startswith('ishape'))):
+                        sel[np.isnan(c[key])] = 0
```
carlosggarcia, line 111: You say in the previous comment that you wan to keep the NaNs. However, I think the elements in `sel` corresponding to NaNs are set to 0, and then will be removed by L.113. What am I missing?

xcell/mappers/mapper_HSC_DR1wl.py
```diff
@@ -0,0 +1,278 @@
+from .mapper_base import MapperBase
+from .utils import get_map_from_points
+from astropy.table import Table, vstack
+import os
+import numpy as np
+import healpy as hp
+
+
+class MapperHSCDR1wl(MapperBase):
```
carlosggarcia, line 9: Could you add some documentation? At least the arguments to pass to the mapper and its default paths/values

xcell/mappers/mapper_HSC_DR1wl.py
```diff
@@ -0,0 +1,278 @@
...
+                                                   cat_cosmos['S17a_objid'],
+                                                   return_indices=True)
+                cat_photo = cat_photo[id_ph]
+                cat_cosmos = cat_cosmos[id_cs]
+
+                z0, zf = self.z_edges
+                msk = np.where((cat_photo['PHOTOZ_BEST'] <= zf) &
+                               (cat_photo['PHOTOZ_BEST'] > z0))[0]
+                cosmos_masked = cat_cosmos[msk]
+                w = cosmos_masked['SOM_weight']*cosmos_masked['weight_source']
```
carlosggarcia, line 257: Could you add a comment here saying that since this is shear, you have to multiply the shear weights to the DIR weights?

xcell/mappers/mapper_HSC_DR1wl.py
```diff
@@ -0,0 +1,278 @@
...
+            # Collect sample cuts
+            sel_area = c['wl_fulldepth_fullcolor']
+            sel_clean = sel_area & c['clean_photometry']
+            sel_maglim = np.ones(len(c), dtype=bool)
+            sel_maglim[c['icmodel_mag'] -
+                       c['a_i'] > self.icut] = 0
+            # Blending
+            sel_blended = np.ones(len(c), dtype=bool)
+            # abs_flux<10^-0.375
+            sel_blended[c['iblendedness_abs_flux'] >= 0.42169650342] = 0
```
carlosggarcia, line 124: It'd be great to have a comment here with the reference to where these cuts come from. I couldn't find the numbers in the HSC paper in a quick search.

xcell/tests/test_mapper_HSCDR1wl.py
```diff
@@ -0,0 +1,211 @@
...
+    for fname in [f'{bn}.fits',
+                  f'{bn}_e1_ns{nside}.fits.gz',
+                  f'{bn}_e2_ns{nside}.fits.gz',
+                  f'{bn}_mask_ns{nside}.fits.gz',
+                  f'{bn}_w2s2_ns{nside}.fits.gz',
+                  f'{bn}_nz.npz']:
+        fname_full = os.path.join(plite, "HSCDR1wl_" + fname)
+        assert os.path.isfile(fname_full)
+
+    # Check we recover the same mas and catalog
```
carlosggarcia, line 137: typo: mas -> maps

xcell/tests/test_mapper_HSCDR1wl.py
```diff
@@ -0,0 +1,211 @@
+import numpy as np
+import xcell as xc
+import healpy as hp
+import os
+import glob
+from astropy.table import Table
+
+
+def make_hsc_nz_data():
```
carlosggarcia, line 9: Are we going to start writing the surveys test data on the fly? Or is this just for HSC because it's a bit more complicated?

carlosggarcia (changes requested):
Some minor comments and questions

xcell/mappers/mapper_HSC_DR1wl.py
```diff
@@ -0,0 +1,278 @@
...
+        self.w_name = 'ishape_hsm_regauss_derived_shape_weight'
+        self.npix = hp.nside2npix(self.nside)
+
+        self.nl_coupled = None
+        self.dndz = None
+        self.cat = None
+        self.mask = None
+        self.signal_map = None
+
+    def _check_lite_exists(self, file_name):
```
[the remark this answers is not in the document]
damonge: Just opened an issue for this

xcell/mappers/mapper_HSC_DR1wl.py
```diff
@@ -0,0 +1,278 @@
...
+                shearmask = ishape_flags_mask *\
+                    ishape_sigma_mask *\
+                    ishape_res_mask *\
+                    ishape_shear_mod_mask*star_mask *\
+                    fdfc_mask
+                cat = cat[shearmask]
+
+                # Redshift bin cut
+                zs = cat['pz_best_eab']
+                zbinmask = (zs <= self.z_edges[1]) & (zs > self.z_edges[0])
```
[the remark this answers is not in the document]
damonge: It is large, but not large enough that it's worth saving yet another intermediate catalog (the redshift-sliced lite catalogs are already pretty small)

xcell/mappers/mapper_HSC_DR1wl.py
damonge: 0 is the equivalent of False if the array is already of type bool (which it is)

xcell/mappers/mapper_HSC_DR1wl.py
damonge: What this does is, if a nan value is found for any property of a galaxy that is not photo-z or ishape, then remove that galaxy

xcell/tests/test_mapper_HSCDR1wl.py
damonge: HSC has sooo many columns (because we're starting from the raw catalogs) that creating a catalog and storing it in the repo will make the size of the repo too large. I would actually advocate that we start doing this for all catalogs in the future (and even do it for the previous datasets if we ever have the spare time)

xcell/mappers/mapper_HSC_DR1wl.py
damonge: Added docs for the different arguments. Better docs are needed for all mappers (and this is already in #73 )

xcell/mappers/mapper_HSC_DR1wl.py
damonge: done

xcell/mappers/mapper_HSC_DR1wl.py
damonge: Added

xcell/tests/test_mapper_HSCDR1wl.py
damonge: done

damonge:
Thanks a lot. Back to you @carlosggarcia !

xcell/mappers/mapper_HSC_DR1wl.py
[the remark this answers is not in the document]
carlosggarcia: Great! Thanks!

xcell/tests/test_mapper_HSCDR1wl.py
carlosggarcia: makes sense. I would also be happy with that approach

xcell/mappers/mapper_HSC_DR1wl.py
carlosggarcia: Could you add the default values in the docstring? E.g. `'depth_cut' : i-band magnitude cut (default 24.5)` or sth like that

xcell/mappers/mapper_HSC_DR1wl.py
```diff
@@ -0,0 +1,293 @@
...
+            sel_fluxcut_y = np.ones(len(c), dtype=int)
+            sel_fluxcut_y[c['ycmodel_flux'] < 5*c['ycmodel_flux_err']] = 0
+            # S/N in grzy (at least 2 pass)
+            sel_fluxcut_grzy = (sel_fluxcut_g+sel_fluxcut_r +
+                                sel_fluxcut_z+sel_fluxcut_y >= 2)
+            # Overall S/N
+            sel_fluxcut = sel_fluxcut_i*sel_fluxcut_grzy
+            # Stars
+            sel_stars = np.ones(len(c), dtype=bool)
+            sel_stars[c['iclassification_extendedness'] > 0.99] = 0
```
carlosggarcia, line 161: I think stars are `iclassification_extendedness = 0` (last line of p.16 in the paper you cite above). Why do you use 0.99?

xcell/mappers/mapper_HSC_DR1wl.py
```diff
@@ -0,0 +1,293 @@
...
+            # Galaxies
+            sel_gals = np.ones(len(c), dtype=bool)
+            sel_gals[c['iclassification_extendedness'] < 0.99] = 0
```
carlosggarcia, line 164: Not sure where this comes from, either

xcell/mappers/mapper_HSC_DR1wl.py
carlosggarcia: Thanks!

carlosggarcia (changes requested):
Last minor comments!

xcell/mappers/mapper_HSC_DR1wl.py
damonge: right now this is a binary value (0 or 1), so it doesn't really matter, but in future data releases this will be a continuous variable and one needs to cut at some point. 0.99 is just a reminder of that (and right now it doesn't matter)

xcell/mappers/mapper_HSC_DR1wl.py
damonge: see above

xcell/mappers/mapper_HSC_DR1wl.py
damonge: Done (BTW, if you mark comments as resolved I may not see them)

carlosggarcia (approved):
LGTM

PR #171, opened 2021-11-12
Adding CatWISE dataset to repo

oliveirafrancofelipe:


xcell/mappers/mapper_CatWISE.py
```diff
@@ -0,0 +1,117 @@
...
+        config - dict
+          {'data_catalog': 'xcell/tests/data/
+          catwise_agns_masked_final_w1lt16p5_alpha.fits',
+           'mask': 'xcell/tests/data/MASKS_exclude_master_final.fits',
+           'mask_name': 'mask_CatWISE'}
+        """
+        self._get_defaults(config)
+        self.file_sourcemask = config.get('mask_sources', None)
+        self.ra_name = 'ra'
+        self.dec_name = 'dec'
```
damonge, line 20: do we need to store these two (self.ra_name, self.dec_name) for any reason?

xcell/mappers/mapper_CatWISE.py
```diff
@@ -0,0 +1,117 @@
...
+        self.radius_name = 'radius'
```
damonge, line 21: same question about `self.radius_name`

xcell/mappers/mapper_CatWISE.py
```diff
@@ -0,0 +1,117 @@
...
+            GLON, GLAT = r(self.cat_data['ra'], self.cat_data['dec'],
+                           lonlat=True)
+            self.cat_data['GLON'] = GLON
+            self.cat_data['GLAT'] = GLAT
+            # Angular and flux conditions
+            self.cat_data = self.cat_data[
+                (self.cat_data['w1'] <
+                 self.config.get('flux_max_W1', 16.4)) &
+                (np.fabs(self.cat_data['GLAT']) >
+                 self.config.get('GLAT_max_deg', 30))]
```
damonge, line 48: The catalog seems to be already cut at `GLAT > 30`, so we don't need to include this

xcell/tests/test_mapper_CatWISE.py
```diff
@@ -0,0 +1,88 @@
+import numpy as np
+import xcell as xc
+import healpy as hp
+import os
+from astropy.table import Table
+
+
```
damonge, line 7: general question about the tests: are you ever testing the lines of code that mask the circles around the sources? For some reason coveralls isn't really complaining, but I can't see where that's happening...

damonge (changes requested):
Looks great, good job @oliveirafrancofelipe ! A few minor comments

xcell/mappers/mapper_CatWISE.py
oliveirafrancofelipe: actually we don't

xcell/mappers/mapper_CatWISE.py
oliveirafrancofelipe: Yes, this is true. The only condition necessary is the flux for W1

xcell/tests/test_mapper_CatWISE.py
oliveirafrancofelipe: No, I didn't write a test for it. In the test for get_mask I've used `config.pop('mask_sources')` so neglecting the holes. Should I include a test for it?

xcell/mappers/mapper_CatWISE.py
```diff
@@ -0,0 +1,112 @@
...
+    def get_catalog(self):
+        if self.cat_data is None:
+            file_data = self.config['data_catalog']
+            self.cat_data = Table.read(file_data)
+            # Galactic coordinates
+            r = hp.Rotator(coord=['C', 'G'])
+            GLON, GLAT = r(self.cat_data['ra'], self.cat_data['dec'],
+                           lonlat=True)
+            self.cat_data['GLON'] = GLON
+            self.cat_data['GLAT'] = GLAT
```
damonge, line 39: Don't calculate galactic coordinates if you're not gonna use them.

xcell/tests/test_mapper_CatWISE.py
damonge: ok, let's move on without this test and circle back if we need it.

damonge (commented):
OK, last comment. We should think about what to do regarding the redshift distribution. Can you create a placeholder method that reads:
```
    def get_nz(self, dz=0):
        raise NotImplementedError("No dNdz for CatWISE yet")
```
?

damonge (approved):
LGTM!

PR #173, opened 2021-11-16
new branch to avoid conflicts

JaimeRZP:


xcell/mappers/mapper_Planck_base.py
```diff
@@ -25,27 +25,45 @@ def _get_Planck_defaults(self, config):
         self.cls_cov = None
         self.custom_auto = True
         self.mask = None
+        self.gal_mask = None
```
damonge: these two (gal_mask and ps_mask) are not needed as attributes

damonge (approved):
LGTM

Issue #185, opened 2022-01-25
Parallel run: check if other process has started writing one of the output files


carlosggarcia:
When launching all runs in parallel, sometimes two processes try to write the workspace at the same time. That makes them fail. We should recheck if the file exists before trying to write. Ideally, we would want just one of the processes do the work and the other wait to read the file, but that's much more effort.

carlosggarcia:
@damonge are you happy to close this issue?

damonge:
yep

PR #191, opened 2022-02-14
Performance

### LINKED ISSUE #75: cls: Improve performance by reducing the size of data read from disk
One of the main bottle necks at this moment is the loading of the (cov)workspaces. We do not need all the information stored in them. We should only load what we actually need by accessing the corresponding part of the fits file.

carlosggarcia:
Closes #75

### DIFF
--- a/xcell/cls/cl.py
+++ b/xcell/cls/cl.py
@@ -121,22 +121,22 @@ class Cl(ClBase):
         f2 = mapper2.get_nmt_field()
         return f1, f2
 
-    def get_workspace(self):
+    def get_workspace(self, read_unbinned_MCM=True):
         if self._w is None:
-            self._w = self._compute_workspace()
+            self._w = self._compute_workspace(read_unbinned_MCM)
         return self._w
 
     def get_workspace_cov(self):
         if self._wcov is None:
             spin0 = self.data.data['cov'].get('spin0', False)
             if spin0 and (self.get_spins() != (0, 0)):
-                self._wcov = self._compute_workspace(spin0=spin0)
+                self._wcov = self._compute_workspace(spin0=spin0, read_unbinned_MCM=False)
             else:
-                self._wcov = self.get_workspace()
+                self._wcov = self.get_workspace(read_unbinned_MCM=False)
 
         return self._wcov
 
-    def _compute_workspace(self, spin0=False):
+    def _compute_workspace(self, spin0=False, read_unbinned_MCM=True):
         # Check if the fields are already of spin0 to avoid computing the
         # workspace twice
         spin0 = spin0 and (self.get_spins() != (0, 0))
@@ -165,7 +165,7 @@ class Cl(ClBase):
             w.write_to(fname)
             self.recompute_mcmc = False
         else:
-            w.read_from(fname)
+            w.read_from(fname, read_unbinned_MCM)
         return w
 
     def get_cl_file(self):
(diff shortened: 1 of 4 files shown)


carlosggarcia:
@damonge you need to update pymaster (or we have to change the requierements to point to the correct branch). What do you think?

damonge:
Blargh. Would it be enough to pull it from master? I'd rather not make a release for now.

carlosggarcia:
@damonge done.

carlosggarcia:
The last commit would close #185. It only checks the workspaces, not the npz because the latter can be overwritten without raising error (although not sure if the file could be corrupted if two processes try to write at the same time).

xcell/cls/cl.py
```diff
@@ -112,22 +121,24 @@ def get_nmt_fields(self):
...
+                self._compute_workspace(read_unbinned_MCM=read_unbinned_MCM)
         return self._w
 
     def get_workspace_cov(self):
         if self._wcov is None:
             spin0 = self.data.data['cov'].get('spin0', False)
             if spin0 and (self.get_spins() != (0, 0)):
-                self._wcov = self._compute_workspace(spin0=spin0)
+                self._wcov = self._compute_workspace(spin0=spin0,
+                                                     read_unbinned_MCM=False)
```
damonge, line 135: why do we call `_compute_workspace` here instead of `get_workspace` as we do below?

xcell/cls/cov.py
```diff
@@ -5,6 +5,7 @@
 import os
 import numpy as np
 import pymaster as nmt
+import time
```
damonge, line 8: do we need this anymore?

xcell/cls/cov.py
```diff
@@ -481,9 +510,18 @@ def get_covariance(self):
...
+        ftime = time.time()
+        print('Added all covariances terms. It took ' +
+              f'{(ftime - itime) / 60} min', flush=True)
+
+        itime = time.time()
         np.savez_compressed(fname, cov=self.cov, cov_G=cov_G, cov_NG=cov_NG,
                             cov_nl_marg=cov_nlm, cov_m_marg=cov_mm)
+        ftime = time.time()
+        print(f'Saved cov npz file. It took {(ftime - itime) / 60} min',
+              flush=True)
```
damonge, line 524: Do we want to remove all the timings?

damonge (commented):
Small comments

xcell/cls/cl.py
carlosggarcia: Because `get_workspace` is the method to get `self._w`.

xcell/cls/cov.py
carlosggarcia: I wanted to know your thoughts about it. I think it might be interesting to know the numbers in the future. Leaving it here would save us time if we later want to do the timing again for some reason with negligible impact on the performance.

xcell/cls/cov.py
carlosggarcia: See previous comment

xcell/cls/cl.py
damonge: Wait, we're calling `get_workspace` just below to get `self._wcov`

damonge: Oh, wait, `get_workspace` doesn't have a `spin0` option. I think I get it.

damonge (approved):

xcell/cls/cl.py
carlosggarcia: Exactly. You use get_workspace in that case because you are not using the spin0 approx.

PR #192, opened 2022-02-16
np_arrays+yml names

JaimeRZP:
all signal_maps homogenized and made into arrays also changed names of yml files


carlosggarcia (changes requested):
The mapper classes within the yml files have to be updated, not only the filenames! For instance, the class `MappereBOSSQSO` does not longer exist.

I haven't checked all the other mappers, but the changes to the signal_map looked good to me. I might have a closer look later.

JaimeRZP:
I just checked. Solved the eBOSS also P18CMBk --> P18CMBK

PR #193, opened 2022-02-16
Shot noise

JaimeRZP:
tentative functions still have to figure out where to plug this


damonge:
Closing since we decided not to support this

PR #201, opened 2022-03-08
Response to damonge's comments -- reg CatWISE

svonhausegger:


svonhausegger:
The keyword `apply_ecliptic_correction` is unique to `get_signal_map()` in `mapper_CatWISE.py`, as (so far) no other catalog requires changes in number density.  You'll probably have to consider whether this keyword should remain globally accessible.  
The second assertion in (at least) `test_get_signal_map()` in `test_mapper_CatWISE.py` fails due to this correction.  For now I therefore propose to set `apply_ecliptic_correction=False`.

damonge:
OK, so about the ecliptic correction, how about the following:
- Make it an option in the config file. If not present, the default is `True`.
- Then, we remove the argument from `get_signal_map`, and instead do `if self.config.get('ecliptic_correction', True):`.

Then, it would be good to have a test for this part of the code (make sure it's doing what it's supposed to do)

xcell/mappers/mapper_CatWISE.py
```diff
@@ -37,17 +40,39 @@ def get_catalog(self):
                  self.config.get('flux_max_W1', 16.4))]
         return self.cat_data
 
+    # Correction to Density
+    def get_ecliptic_correction(self):
```
JaimeRZP, line 44: I believe it would be good if we make get_eclipttic_correction a private method.
`get_ecliptic_correction(self) --> _get_ecliptic_correction(self)`

xcell/mappers/mapper_CatWISE.py
```diff
@@ -37,17 +40,39 @@ def get_catalog(self):
...
+    def get_signal_map(self):
         if self.delta_map is None:
             d = np.zeros(self.npix)
             self.cat_data = self.get_catalog()
             self.mask = self.get_mask()
+            # ecliptic latitude correction -- SvH 5/3/22
+            if self.apply_ecliptic_correction:
+                correction = self.get_ecliptic_correction()
+            else:
+                correction = np.zeros_like(d)
```
JaimeRZP, line 68: For readability I would move this step after calculating the `nmap` ; i.e. just before line 72.
Also `correction = self.get_ecliptic_correction()  --> correction = self._get_ecliptic_correction()`

xcell/tests/test_mapper_CatWISE.py
```diff
@@ -60,6 +61,17 @@ def test_get_signal_map():
     clean_fake_data()
 
 
+def test_ecliptic_correction():
+    config = get_config()
+    m = xc.mappers.MapperCatWISE(config)
+    d = m.get_ecliptic_correction()
```
JaimeRZP, line 67: As before: `m.get_ecliptic_correction() --> m._get_ecliptic_correction()`

JaimeRZP (changes requested):
Very minor comments. 
Mainly making the new method private (adding a '_') as I don't believe users should have access to it.

JaimeRZP (approved):

damonge:
@JaimeRZP @svonhausegger you probably need to merge in the latest version of the master branch. @carlosggarcia fixed this in this PR https://github.com/xC-ell/xCell/pull/205

PR #202, opened 2022-03-14
Loading parts of the master yml file from other yml files

damonge:
E.g. the yaml file now can be:

```
mappers:
  mapper1: !include path/to/mapper1.yml
  mapper2: !include path/to/mapper1.yml
  mapper2: !include path/to/mapper1.yml

...
```


carlosggarcia (approved):
LGTM. Thanks for handling it!

PR #215, opened 2022-04-11
All mappers save masks

### LINKED ISSUE #214: Should all mappers have rerun options for masks?
Fact 1: a majority of our catalog-based mappers use equatorial coordinates. Most map-based mappers use galactic coordinates.
Fact 2: it is more precise to rotate the coordinates and ellipticities of a catalog than to rotate a map (the latter involves interpolation or lossy spherical harmonic transforms).
Fact 3: map rotations are kind of slow.
Conclusion: it will be best to use galactic coordinates as a basis for our power spectra. However, this means that we will have to rotate the masks of all mappers using equatorial coordinates. Since this is a bit slow, we could save those masks. Since many of these masks are binary, saving them in `fits.gz` format will probably be quite efficient.

Question 1: should all mappers have the possibility of saving rerun versions of their masks?
Question 2: should we make the rerun path a global parameter, so all rerun files are dumped in the same place?

This would have the added advantages of:
a) Faster C_ells and covariances (since masks are computed every time)
b) Homogeneous naming scheme for rerun files.

damonge:
Closes #214

### DIFF
--- a/xcell/cls/data.py
+++ b/xcell/cls/data.py
@@ -284,6 +284,9 @@ class Data():
         config = self.data['tracers'][tr]
         nside = self.data['sphere']['nside']
         coords = self.data['sphere']['coords']
+        config['ignore_rerun'] = config.get('ignore_rerun',
+                                            self.data.get('ignore_rerun',
+                                                          False))
         if 'nside' not in config:
             config['nside'] = nside
         elif config['nside'] != nside:
--- a/xcell/mappers/mapper_KV450.py
+++ b/xcell/mappers/mapper_KV450.py
@@ -53,7 +53,6 @@ class MapperKV450(MapperBase):
         self.signal_map = None
         self.maps = {'PSF': None, 'shear': None, 'stars': None}
 
-        self.mask = None
         self.masks = {'stars': None, 'galaxies': None}
 
         self.nl_coupled = None
@@ -186,7 +185,7 @@ class MapperKV450(MapperBase):
                                       rot=self.rot)
             return msk
 
-        fn = '_'.join([f'KV450_mask_{kind}_bin{self.zbin}',
+        fn = '_'.join([f'mask_{self.mask_name}_{kind}',
                        f'coord{self.coords}',
                        f'ns{self.nside}.fits.gz'])
         self.masks[kind] = self._rerun_read_cycle(fn, 'FITSMap',
--- a/xcell/mappers/mapper_KV450.py
+++ b/xcell/mappers/mapper_KV450.py
@@ -173,25 +173,16 @@ class MapperKV450(MapperBase):
     def get_mask(self, mode=None):
         kind, e1f, e2f, mod = self._set_mode(mode)
         if self.masks[kind] is not None:
-            self.mask = self.masks[kind]
-            return self.mask
+            return self.masks[kind]
 
-        def get_mask_mod():
-            data = self._get_gals_or_stars(kind)
-            msk = get_map_from_points(data, self.nside,
-                                      w=data['weight'],
-                                      ra_name='ALPHA_J2000',
-                                      dec_name='DELTA_J2000',
-                                      rot=self.rot)
-            return msk
-
-        fn = '_'.join([f'mask_{self.mask_name}_{kind}',
-                       f'coord{self.coords}',
-                       f'ns{self.nside}.fits.gz'])
-        self.masks[kind] = self._rerun_read_cycle(fn, 'FITSMap',
-                                                  get_mask_mod)
-        self.mask = self.masks[kind]
-        return self.mask
+        data = self._get_gals_or_stars(kind)
+        msk = get_map_from_points(data, self.nside,
+                                  w=data['weight'],
+                                  ra_name='ALPHA_J2000',
+                                  dec_name='DELTA_J2000',
+                                  rot=self.rot)
+        self.masks[kind] = msk
+        return msk
 
     def _get_w2s2(self, mode):
         kind, e1f, e2f, mod = self._set_mode(mode)
--- a/xcell/mappers/mapper_KV450.py
+++ b/xcell/mappers/mapper_KV450.py
@@ -88,9 +88,8 @@ class MapperKV450(MapperBase):
             save_rerun_data(self, fn, 'FITSTable', cat.as_array())
         return cat_bins[self.zbin].as_array()
 
-    def _set_mode(self, mode=None):
-        if mode is None:
-            mode = self.mode
+    def _set_mode(self):
+        mode = self.mode
 
         if mode == 'shear':
             kind = 'galaxies'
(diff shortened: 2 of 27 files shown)


xcell/cls/data.py
```diff
@@ -284,6 +284,9 @@ def get_mapper(self, tr):
         config = self.data['tracers'][tr]
         nside = self.data['sphere']['nside']
         coords = self.data['sphere']['coords']
+        config['ignore_rerun'] = config.get('ignore_rerun',
+                                            self.data.get('ignore_rerun',
+                                                          False))
```
carlosggarcia, line 289: Hmm, is this a global variable? I think it could also be helpful to have it defined per tracer (imagine that you change a mapper which modifies the mask you will only want to ignore the rerun for that case). Also, will `ignore_rerun` overwrite what you have?

xcell/mappers/mapper_base.py
```diff
@@ -37,6 +38,14 @@ def get_contaminants(self):
         return None
 
     def get_mask(self):
+        if self.mask is None:
+            fn = '_'.join([f'mask_{self.mask_name}',
```
carlosggarcia, line 42: Hmm, do you need a name for the mask? Why not the same name as for the other maps? (i.e. `self.map_name`)?

xcell/mappers/mapper_KV450.py
```diff
@@ -186,7 +185,7 @@ def get_mask_mod():
                                       rot=self.rot)
             return msk
 
-        fn = '_'.join([f'KV450_mask_{kind}_bin{self.zbin}',
+        fn = '_'.join([f'mask_{self.mask_name}_{kind}',
```
carlosggarcia, line 188: Why do you keep this here for KV450?

xcell/mappers/mapper_KiDS1000.py
```diff
@@ -176,7 +175,7 @@ def get_mask_mod():
                                       rot=self.rot)
             return msk
 
-        fn = '_'.join([f'KiDS1000_mask_{kind}_bin{self.zbin}',
+        fn = '_'.join([f'mask_{self.mask_name}_{kind}',
```
carlosggarcia, line 178: Same as for KV450

xcell/mappers/mapper_SDSS.py
```diff
@@ -94,7 +88,9 @@ def _get_signal_map(self):
 
     def get_signal_map(self):
         if self.signal_map is None:
-            fn = self._get_map_fname(mp='signal')
+            fn = '_'.join([f'SDSS_{self.SDSS_name}_signal',
```
carlosggarcia, line 91: I kind of prefer to have fixed names for the mappers like here instead of relying on the user

xcell/mappers/mapper_base.py
carlosggarcia: In fact, if you like that, I think we can give a `self.map_name` to each mapper that it's fixed (or a default that can be changed)

carlosggarcia (changes requested):
Looks good. Two comments:
 - I think we should have a default map_name for each mapper.
 - KV450 and K1000 don't follow the others mappers for the masks. I would prefer this to be unified if it is not too much work.
 - Some tests have failed. Check them. (Haven't checked the tests too carefully, but they looked good)

xcell/cls/data.py
damonge: This allows you to have it defined per tracer, and if it is, then that overrides the global variable.

xcell/mappers/mapper_base.py
damonge: In other words, all tracers need a `mask_name` (otherwise initialization fails)

xcell/mappers/mapper_KV450.py
damonge: KiDS and KV450 can generate masks not just for the catalog but for the star subsample.
If you want them to adhere to the rest of the mappers, we would need to drop that functionality.

xcell/mappers/mapper_KiDS1000.py
damonge: See above

xcell/mappers/mapper_SDSS.py
damonge: We only have this structure for SDSS and ACT. Doing this would be dangerous, though. For example, the `MapperBOSS` class can be either the CMASS or LOWZ samples, and those should have different names.

Moreover, imagine you want to divide them into different z-bins, so you could have `CMASS1`, `CMASS2` etc.

I think it's best to rely on the user.

damonge: (which is us)

xcell/cls/data.py
carlosggarcia: Right. My brain ignored the `config.get` part...

xcell/mappers/mapper_KV450.py
carlosggarcia: But not in the same run, right? If you want to keep this interactive functionality, you could add `**kwargs` to `get_mask` and pass them `_rerun_read_cycle` and to `_get_mask`.

damonge: you could generate them in the same run. You just need to add two mappers with different `mode` values in the config dictionary.

I think it's not worth the complication. To some extent this is what overriding is for

damonge: ... unless we want to assume that users (again, us) would be attentive enough to pass different mask names to both yaml sections... If you're happy to assume that (at the end of the day we are assuming that for all other mappers), then this could be incorporated into the general structure.

carlosggarcia: Yes. I would rather unify things. And as you say, it is reasonable to assume the user will choose the right mask name.

damonge: OK, done. Take a look!

xcell/mappers/mapper_KV450.py
```diff
@@ -174,25 +173,16 @@ def get_ellip_maps_mod():
     def get_mask(self, mode=None):
```
carlosggarcia, line 173: Shouldn't it be `_get_mask`? (with the underscore)

xcell/mappers/mapper_KiDS1000.py
```diff
@@ -79,9 +78,8 @@ def _load_catalog(self):
                 cat_out = cat
         return cat_out
 
-    def _set_mode(self, mode=None):
-        if mode is None:
-            mode = self.mode
+    def _set_mode(self):
```
carlosggarcia, line 81: Why do you remove the `mode` argument? (or why do you remove it here and keep it in KV450?)

carlosggarcia (changes requested):
Minor comments. Also, doesn't `test_mapper_kv450` need to be updated?

xcell/mappers/mapper_KV450.py
damonge: you're right!

xcell/mappers/mapper_KiDS1000.py
damonge: The logic is that the mode is set in the yml file, and cannot be changed afterwards

carlosggarcia: Make sense (and I like this more)

carlosggarcia (approved):
LGTM! Thanks, @damonge!

PR #222, opened 2022-04-25
Super Sample Covariance implementation

carlosggarcia:
@damonge I have basically reused what we had for non-Gaussian covariances. 

A few questions:
- Do we want it split from the nGC?
- The different nGC terms are computed in a for loop that calls `get_covariance_ng_halomodel` for each specific type. In this functions, we are regenerating the `ccl_tracers`in each call. Shouldn't we store this and only compute it once? 
- when calling `ccl.angular_cl_cov_SSC`, I'm not passing `sigma2_B`. Should I pass one with the same `a_arr` as for the `Tkk`? (it's commented out now in the code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        py: [ 3.7 ]
+        py: [ 3.6 ]
         CC: [ gcc ]
 
     steps:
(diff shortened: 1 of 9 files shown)


carlosggarcia:
@damonge I've finished the implementation of the SSC with the linear bias approximation. For the test to work I had to use exactly the same array of scale factors to compute the sigma2_B. I do not understand this and I'm fearing I might have messed something up, if it is what it is or if it is a bug in CCL. It'd be great if you have a look.

carlosggarcia:
Implemented in #305

PR #239, opened 2022-09-19
Run covariance blocks in batches

carlosggarcia:
Improved `run_cls.py` to allow the computation of the covariance blocks in batches sharing the same covariance workspace.

In order to use it one has to run it with `--batches`. In addition, if one wants to remove the covariance workspace once the job has finished, one can use the argument `--remove_cwsp` together with `--batches`.


damonge (approved):
I think it looks good. However, note that in principle we get O(half a million) covariances, so it might be good to check if the `get_jobs_with_same_cwsp` function actually runs on the master yaml file.

carlosggarcia:
You were right. I stopped after 2h of running. I'll think of sth better

carlosggarcia:
@damonge back to you. This should work

PR #240, opened 2022-09-19
Add get_signal_map in base class + auto save maps

### LINKED ISSUE #236: Planck mapper rotation
We should save Planck maps after rotating them, since this takes a long time. This only needs to be done when using Galactic coords for the Planck maps, or when using Galactic coords for the SPT map.

carlosggarcia:
Closes #236. 

@damonge if you like the approach, I will modify the other mappers accordingly

### DIFF
--- a/input/CIBLenz__353.yml
+++ b/input/CIBLenz__353.yml
@@ -1,4 +1,5 @@
 mapper_class: MapperCIBLenz
+band: 353
 file_map: /mnt/extraspace/damonge/Datasets/CIB_Lenz/353/2.5e+20_gp20/cib_fullmission.hpx.fits
 file_hm1: /mnt/extraspace/damonge/Datasets/CIB_Lenz/353/2.5e+20_gp20/cib_evenring.hpx.fits
 file_hm2: /mnt/extraspace/damonge/Datasets/CIB_Lenz/353/2.5e+20_gp20/cib_oddring.hpx.fits
--- a/xcell/mappers/mapper_2MPZ.py
+++ b/xcell/mappers/mapper_2MPZ.py
@@ -26,7 +26,6 @@ class Mapper2MPZ(MapperBase):
         self.npix = hp.nside2npix(self.nside)
 
         # Angular mask
-        self.dndz = None
         self.nl_coupled = None
 
     def _get_coords(self):
--- a/xcell/mappers/mapper_2MPZ.py
+++ b/xcell/mappers/mapper_2MPZ.py
@@ -7,6 +7,8 @@ import os
 
 
 class Mapper2MPZ(MapperBase):
+    map_name = '2MPZ'
+
     def __init__(self, config):
         """
         config - dict
@@ -25,7 +27,7 @@ class Mapper2MPZ(MapperBase):
 
         # Angular mask
         self.dndz = None
-        self.delta_map = None
+        self.signal_map = None
         self.nl_coupled = None
 
     def _get_coords(self):
@@ -92,20 +94,19 @@ class Mapper2MPZ(MapperBase):
             self.dndz = self._rerun_read_cycle(fn, 'NPZ', self._get_nz)
         return self._get_shifted_nz(dz, return_jk_error=return_jk_error)
 
-    def get_signal_map(self):
-        if self.delta_map is None:
-            d = np.zeros(self.npix)
-            self.cat_data = self.get_catalog()
-            self.mask = self.get_mask()
-            nmap_data = get_map_from_points(self.cat_data, self.nside,
-                                            ra_name=self.ra_name,
-                                            dec_name=self.dec_name)
-            mean_n = np.average(nmap_data, weights=self.mask)
-            goodpix = self.mask > 0
-            # Division by mask not really necessary, since it's binary.
-            d[goodpix] = nmap_data[goodpix]/(mean_n*self.mask[goodpix])-1
-            self.delta_map = np.array([d])
-        return self.delta_map
+    def _get_signal_map(self):
+        d = np.zeros(self.npix)
+        self.cat_data = self.get_catalog()
+        self.mask = self.get_mask()
+        nmap_data = get_map_from_points(self.cat_data, self.nside,
+                                        ra_name=self.ra_name,
+                                        dec_name=self.dec_name)
+        mean_n = np.average(nmap_data, weights=self.mask)
+        goodpix = self.mask > 0
+        # Division by mask not really necessary, since it's binary.
+        d[goodpix] = nmap_data[goodpix]/(mean_n*self.mask[goodpix])-1
+        signal_map = np.array([d])
+        return signal_map
 
     def _get_mask(self):
         # We will assume the mask has been provided in the right
--- a/xcell/mappers/mapper_2MPZ.py
+++ b/xcell/mappers/mapper_2MPZ.py
@@ -27,7 +27,6 @@ class Mapper2MPZ(MapperBase):
 
         # Angular mask
         self.dndz = None
-        self.signal_map = None
         self.nl_coupled = None
 
     def _get_coords(self):
--- a/xcell/mappers/mapper_ACT_compsept.py
+++ b/xcell/mappers/mapper_ACT_compsept.py
@@ -20,14 +20,13 @@ class MapperACTCompSept(MapperACTBase):
         self._get_ACT_defaults(config)
 
     def _get_signal_map(self):
-        if self.signal_map is None:
-            # The 'Weights' FITS file contains the 2D Fourier space
-            # weight for each pixel corresponding to the detector array.
-            signal_map = enmap.read_map(self.file_map)
-            signal_map = reproject.healpix_from_enmap(signal_map,
-                                                      lmax=self.lmax,
-                                                      nside=self.nside)
-            signal_map = rotate_map(signal_map, self.rot)
+        # The 'Weights' FITS file contains the 2D Fourier space
+        # weight for each pixel corresponding to the detector array.
+        signal_map = enmap.read_map(self.file_map)
+        signal_map = reproject.healpix_from_enmap(signal_map,
+                                                  lmax=self.lmax,
+                                                  nside=self.nside)
+        signal_map = rotate_map(signal_map, self.rot)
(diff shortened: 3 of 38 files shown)


input/CIBLenz__353.yml
```diff
@@ -1,4 +1,5 @@
 mapper_class: MapperCIBLenz
+band: 353
```
damonge, line 2: did we miss this?

xcell/mappers/mapper_2MPZ.py
```diff
@@ -7,6 +7,8 @@
 
 
 class Mapper2MPZ(MapperBase):
+    map_name = '2MPZ'
```
damonge, line 10: Why do we need this?

xcell/mappers/mapper_ACT_compsept.py
```diff
@@ -20,14 +20,13 @@ def __init__(self, config):
         self._get_ACT_defaults(config)
 
     def _get_signal_map(self):
-        if self.signal_map is None:
```
damonge, line 23: this was a huge bug, right? We weren't really updating `self.signal_map`

xcell/mappers/mapper_DESY1gc.py
```diff
@@ -7,6 +7,8 @@
 
 
 class MapperDESY1gc(MapperBase):
+    map_name = 'DESY1gc'
+
```
damonge, line 11: Don't you need to update this name depending on the bin?

xcell/mappers/mapper_DESY1wl.py
```diff
@@ -6,6 +6,8 @@
 
 
 class MapperDESY1wl(MapperBase):
+    map_name = 'DESY1wl'
+
```
damonge, line 10: don't you need to update this name depending on the bin?

xcell/mappers/mapper_NVSS.py
```diff
@@ -23,7 +25,7 @@ def __init__(self, config):
 
         self.npix = hp.nside2npix(self.nside)
         # Angular mask
-        self.delta_map = None
+        self.signal_map = None
```
damonge, line 28: do you need this? Isn't this in base?

xcell/mappers/mapper_Planck_base.py
```diff
@@ -73,9 +70,23 @@ def _get_ps_mask(self):
...
+        if self.hm1_map is None:
+            fn = '_'.join([f'{self.map_name}_hm_maps',
+                           f'coord{self.coords}',
+                           f'ns{self.nside}.fits.gz'])
+
+            hm1_map, hm2_map = self._rerun_read_cycle(fn, 'FITSMap',
+                                                      self._generate_hm_maps)
+
+            self.hm1_map = hm1_map.reshape((1, -1))
+            self.hm2_map = hm2_map.reshape((1, -1))
```
damonge, line 86: Is this `(1, -1)` so that you get a `[1, npix]` array?

xcell/mappers/mapper_ROSAT.py
```diff
@@ -41,7 +43,8 @@ def __init__(self, config):
 
         self.expmap = None
         self.pholist = None
-        self.countrate_map = None
+        # signal_map is a map of the countrate
+        self.signal_map = None
```
damonge, line 47: do you need this here?

xcell/mappers/mapper_WIxSC.py
```diff
@@ -27,12 +29,13 @@ def __init__(self, config):
 
         self.cat_data = None
         self.npix = hp.nside2npix(self.nside)
-        self.bn = self.config['bin_name']
+        # TODO: I think this should be zbin since you're only passing a number
+        # Keeping self.bn for backward compatibility
+        self.zbin = self.bn = self.config['bin_name']
```
damonge, line 34: don't worry about backwards compatibility.

xcell/mappers/mapper_base.py
```diff
@@ -24,6 +27,10 @@ def _get_defaults(self, config):
         self.mask_power = config.get('mask_power', 1)
         self.coords = config['coords']
         self.mask = None
+        self.signal_map = None
+        # Defined here for the signal_map file name
+        self.zbin = None
+        self.dndz = None
```
damonge, line 33: Humm, why do you need `zbin` and `dndz`? These seem too specific to be generally useful.

xcell/mappers/mapper_base.py
```diff
@@ -32,9 +39,31 @@ def _get_rotator(self, coord_default):
...
 
-    def get_signal_map(self):
+    def _get_signal_map(self):
         raise NotImplementedError("Do not use base class")
 
+    def get_signal_map(self, **kwargs):
+        if self.signal_map is None:
+            fn = '_'.join([f'{self.map_name}_signal_map' + (f'_bin{self.zbin}'
+                                                            if self.zbin is not
+                                                            None else ''),
```
damonge, line 49: I'd rather add this to `self.map_name` in the mappers that take redshift bins

xcell/mappers/mapper_base.py
```diff
@@ -32,9 +39,31 @@ def _get_rotator(self, coord_default):
...
+                           *[f'{k}{v}' for k, v in kwargs.items()],
+                           f'coord{self.coords}',
+                           f'ns{self.nside}.fits.gz'])
+
+            print(f'Saving signal_map: {fn}')
```
damonge, line 54: remove this?

xcell/mappers/mapper_base.py
```diff
@@ -32,9 +39,31 @@ def _get_rotator(self, coord_default):
...
+
+            def func():
+                if not kwargs:
+                    return self._get_signal_map()
+                else:
+                    return self._get_signal_map(**kwargs)
+
+            signal_map = self._rerun_read_cycle(fn, 'FITSMap', func)
```
damonge, line 62: How does it know to read 2 maps for spin-2 fields, for instance? Or does that never happen?

damonge (changes requested):
Thanks a lot! A few minor comments.

xcell/mappers/mapper_base.py
carlosggarcia: By default, it loads all columns in the fits file.

damonge: aah, good point

xcell/mappers/mapper_base.py
carlosggarcia: To be able to add them to the file names in the previous format (after 'signal_map')

xcell/mappers/mapper_ROSAT.py
carlosggarcia: Nope. I will remove it

xcell/mappers/mapper_Planck_base.py
carlosggarcia: Yes

xcell/mappers/mapper_NVSS.py
carlosggarcia: to be removed

xcell/mappers/mapper_DESY1wl.py
carlosggarcia: That's an option, but would break the previous file naming convention. Happy to break it though. It would make things easier and cleaner

xcell/mappers/mapper_DESY1gc.py
carlosggarcia: See reply in wl mapper

xcell/mappers/mapper_2MPZ.py
carlosggarcia: To homogenize the  way we save the maps. See the MapperBase class

xcell/mappers/mapper_base.py
damonge: Yes, I see. But I think it's not a huge deal. I'd rather do that than assume that we always have a `zbin`

xcell/mappers/mapper_DESY1wl.py
damonge: do it!

input/CIBLenz__353.yml
carlosggarcia: It wasn't needed because you specified the files to read. I added this to use it to create the map file names

xcell/mappers/mapper_base.py
damonge: I didn't see `dndz` being used

xcell/mappers/mapper_ACT_compsept.py
carlosggarcia: Did you update it? I think it didn't matter because if `self.signal_map` was `None`, `get_signal_map` return `self.signal_map` anyway without calling this function

carlosggarcia:
Replied all comments. I will do the corresponding modifications ASAP. Are we happy to add the bin information then to the `map_name` and break the signal maps file names convention that we had so far?

xcell/mappers/mapper_base.py
carlosggarcia: probably it was another equivalent name to `zbin` that I forgot to remove

carlosggarcia: you have to declare `dndz` because we have the `_get_shifted_nz` method which uses it.

xcell/mappers/mapper_base.py
carlosggarcia: the print statement? I think it's nice to see the path where you're saving things (not strong feelings, though)

carlosggarcia:
@damonge back to you

damonge (changes requested):
If we have 
```
self.signal_map
```
on the base mapper class, this shouldn't appear in any other subclasses.

Same for `self.dndz`.

I think I can see several offenders here:
ROSAT, NVSS, KiDS1000, KV450, DESY1wl, dummy, HSC_DR1wl, DESY1gc, 2MPZ, WIxSC, SDSS, DELS

damonge (approved):

carlosggarcia:
Done. Thanks, @damonge!

PR #247, opened 2022-11-08
Run1

carlosggarcia:


xcell/cls/cl.py
```diff
@@ -323,28 +324,71 @@ def _compute_workspace(self, spin0=False, read_unbinned_MCM=True):
...
+                return w
+
+        n_iter = self.data.data['sphere']['n_iter_mcm']
+        l_toeplitz, l_exact, dl_band = self.data.check_toeplitz('cls')
+        if spin0:
+            m1, m2 = self.get_mappers()
+            msk1 = m1.get_mask()
+            msk2 = m2.get_mask()
+            f1 = nmt.NmtField(msk1, None, spin=0)
+            f2 = nmt.NmtField(msk2, None, spin=0)
```
damonge, line 339: is this simply a reordering of the above, or was there a bug?

xcell/cls/cl.py
```diff
@@ -711,17 +778,18 @@ def get_cl_file(self):
...
             tracers = self.data.data['tracers']
-            fiducial = self.data.data['cov']['fiducial']
             d1, d2 = self.get_dtypes()
-            for dtype, tr in zip([self.tr1, self.tr2], [d1, d2]):
-                if (dtype == 'galaxy_shear') and fiducial['wl_m']:
-                    cl = (1 + tracers[tr]['m']) * cl
+            for tr, dtype in zip([self.tr1, self.tr2], [d1, d2]):
+                if (dtype == 'galaxy_shear'):
+                    m = tracers[tr].get('m', 0)
+                    cl = (1 + m) * cl
```
damonge, line 792: humm, why are we doing this? Is this only for the covariance? Or will you end up storing C_ells in the sacc file that are already corrected for the multiplicative bias?

xcell/cls/cov.py
```diff
@@ -132,44 +134,54 @@ def get_outdir(self):
...
+                                         l_exact=l_exact,
+                                         dl_band=dl_band)
+        if save_cw:
+            tools.save_wsp(cw, fname)
+        self.recompute_cmcm = False
+        self.cw = cw
+
+        return self.cw
+
+    def _get_cl_for_cov(self, clab, clab_fid):
```
damonge, line 184: same question, was this just a reshuffling of the above?

damonge: (beyond the ma, mb thing)

xcell/cls/cl.py
carlosggarcia: It's to avoid crashing the run if the workspace in the disk is corrupted. If after reading the workspace it is corrupted, it will remove the file and compute the workspace again. Before it would have tried to use the corrupted loaded workspace (and fail)

xcell/cls/cl.py
carlosggarcia: This will store fiducial Cells that are corrected for the multiplicative bias. I'm doing this to do the marginalization of `m` correctly.

xcell/cls/cov.py
carlosggarcia: Same as for the workspace. Now we use the `read_wsp` functions that know how to handle the situation of loading a corrupted file.

xcell/cls/cl.py
damonge: so only for the fiducials, not for the final sacc file, right?

damonge (approved):
Assuming the 1+m is not being applied to the final C_ells in the sacc file, I'm happy to merge this

xcell/cls/cl.py
carlosggarcia: Yep. It's inside the `ClFid` class.

carlosggarcia:
Thanks @damonge!! As I said above, yes, the 1+m is inside the `ClFid` class, so it will not affect the data

PR #259, opened 2023-02-06
to_sacc: fixed missing nz

carlosggarcia:
There are galaxy clustering probes we do not know the redshift distribution. That should not prevent us from creating a sacc file.


carlosggarcia:
Merging now. If this breaks sth I will fix it.

PR #261, opened 2023-05-29
GAIA QSO data

damonge:


cosmotheka/mappers/mapper_PR4CMBK.py
```diff
@@ -0,0 +1,43 @@
...
+        #     [l (Array): multipole list,
+        #      Nl (Array): noise power spectrum] (Array)
+
+        if self.noise is None:
+            # Read noise file. Column order is: ['l', 'Nl'], 'Nl+Cl']
+            nl = np.loadtxt(self.config['file_noise'])
+            # Zero nan
+            nl[0] = 0
+            ls = np.arange(len(nl))
+            self.noise = (ls, nl)
```
carlosggarcia, line 41: This is not consistent with `MapperP18CMBK` definition of noise https://github.com/Cosmotheka/Cosmotheka/blob/2ad07fa06e06341c0e411e5652949219efc6de84/cosmotheka/mappers/mapper_P18CMBK.py#L116

Why do you implement a new class instead of modifying the parent class one? I'd do as you did above, and include this nl[0] = 0 in `MapperP18CMBK` method.

cosmotheka/mappers/mapper_Quaia.py
```diff
@@ -0,0 +1,152 @@
...
+        # after binning it in redshift.
+
+        cat = Table.read(self.config['data_catalog'], format='fits')
+        cat.keep_columns(['ra', 'dec', self.z_name, self.z_name+'_err'])
+        z_d = cat[self.z_name]
+        mask_bin = (z_d < self.z_edges[1]) & (z_d >= self.z_edges[0])
+        cat = cat[mask_bin]
+        return cat.as_array()
+
+    def get_catalog(self):
```
carlosggarcia, line 40: Annyoing that we have to define this per mapper...

cosmotheka/mappers/mapper_Quaia.py
carlosggarcia, line 1: I had a quick look and I trust that everything is correct since this was used in two paper already.

cosmotheka/tests/data/catalog.fits
carlosggarcia, line 1: It would be better to generate these on the fly to avoid making the repo too heavy. Is it two slow? Same with the other files.

cosmotheka/tests/test_mapper_p18cmbk.py
```diff
@@ -55,7 +56,7 @@ def test_get_signal_map():
     d = m.get_signal_map()
     assert len(d) == 1
     d = d[0]
-    assert np.all(np.fabs(d-1) < 0.02)
+    assert np.all(np.fabs(d) < 0.02)
```
carlosggarcia, line 59: what has changed? why did `d-1` work before and now you use `d`?

cosmotheka/tests/test_mapper_p18cmbk.py
```diff
@@ -92,5 +93,4 @@ def test_get_nmt_field():
     m = get_mapper()
     f = m.get_nmt_field()
     cl = nmt.compute_coupled_cell(f, f)[0]
-    assert np.fabs(cl[0]-4*np.pi) < 1E-3
```
carlosggarcia, line 95: as above. what changed?

cosmotheka/tests/test_mapper_pr4cmbk.py
carlosggarcia, line 1: not sure you need so many tests given that the class is basically a child of `MapperP18CMBK` with 0 departure from it.

cosmotheka/tests/test_mapper_quaia.py
```diff
@@ -0,0 +1,97 @@
...
+def get_mapper():
+    return xc.mappers.MapperQuaia(get_config())
+
+
+def test_smoke():
+    cleanup_rerun()
+    m = get_mapper()
+    m.get_catalog()
+    assert len(m.cat_data) == hp.nside2npix(32)
+    cleanup_rerun()
```
carlosggarcia, line 38: More elegant solution than adding this at the end of each test is adding this at the beginning of the file

```
@pytest.fixture(autouse=True)
def teardown_test():
    clean_outdir()

def clean_outdir():
    os.system(f"rm -f {OUTDIR}*")
```
or change `clean_outdir` for `cleanup_rerun`

carlosggarcia (changes requested):
Some comments

cosmotheka/mappers/mapper_PR4CMBK.py
```diff
@@ -0,0 +1,43 @@
+from .mapper_P18CMBK import MapperP18CMBK
+import numpy as np
+
+
+class MapperPR4CMBK(MapperP18CMBK):
```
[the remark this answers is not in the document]
damonge: I can add Planck if you want, but not P18

cosmotheka/mappers/mapper_P18CMBK.py
```diff
@@ -43,6 +43,8 @@ def _get_klm(self):
         if self.klm is None:
             # Read alms and rotate if needed
             klm, lmax = hp.read_alm(self.config['file_klm'], return_mmax=True)
+            # First element may be nan. Fix that.
+            klm[0] = 0+0j
```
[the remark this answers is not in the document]
damonge: yes

cosmotheka/mappers/mapper_PR4CMBK.py
damonge: The format of the noise file is completely different for PR4 (enough that writing a new function sounds like a better idea to me)

cosmotheka/mappers/mapper_Quaia.py
damonge: every dataset has its quirks

cosmotheka/tests/test_mapper_p18cmbk.py
damonge: The alms are generated from a constant map=1, so the only non-zero alm is a00. Now we're setting that to zero (as we should), so here we're checking that the recovered map is 0 everywhere.

cosmotheka/tests/test_mapper_p18cmbk.py
damonge: as above

cosmotheka/tests/data/catalog.fits
damonge: I agree. We can open an issue for this if you want. I started doing this for ROSAT, but most of the catalog-based mappers were tested from these precomputed datasets.

cosmotheka/tests/test_mapper_pr4cmbk.py
damonge: Agreed (although they were quite useful to weed out a lot of rubbish from the code). I've reduced them massively.

cosmotheka/mappers/mapper_PR4CMBK.py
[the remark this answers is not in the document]
damonge: (I.e. you tell me what to do)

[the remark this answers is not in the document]
carlosggarcia: Yeah, I would add Planck, I think.

cosmotheka/mappers/mapper_P18CMBK.py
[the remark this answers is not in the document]
carlosggarcia: makes sense

cosmotheka/mappers/mapper_PR4CMBK.py
carlosggarcia: hmm, then, is the comment in L.36 correct? It looks pretty much the same as for `MapperP18CMBK`

cosmotheka/tests/test_mapper_pr4cmbk.py
carlosggarcia: great. thanks!

carlosggarcia (approved):
LGTM. I left a comment regarding the noise. Do as you prefer.

cosmotheka/tests/test_mapper_quaia.py
carlosggarcia: Yep

cosmotheka/mappers/mapper_PR4CMBK.py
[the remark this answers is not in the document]
damonge: great, done

cosmotheka/mappers/mapper_PR4CMBK.py
damonge: you're right, the comment is wrong. I've fixed it.

PR #267, opened 2023-08-15
Inverse-variance mask for ROSAT

damonge:
Using the exposure map itself as a mask for ROSAT (which is similar to inverse-variance noise weighting) leads to more homogeneous noise properties and smaller power spectrum uncertainties.


carlosggarcia (approved):
LGTM

Issue #266, opened 2023-08-14
KiDS1000 does not read saved cat 


carlosggarcia:
KiDS-1000 mapper is not reading the saved cat correctly.

```
KiDS1000_bin0_cat.fits
Saving /mnt/extraspace/damonge/Datasets/KiDS1000/xcell_runs/KiDS1000_bin0_cat.fits
KiDS1000_bin1_cat.fits
Saving /mnt/extraspace/damonge/Datasets/KiDS1000/xcell_runs/KiDS1000_bin1_cat.fits
KiDS1000_bin2_cat.fits
Saving /mnt/extraspace/damonge/Datasets/KiDS1000/xcell_runs/KiDS1000_bin2_cat.fits
KiDS1000_bin3_cat.fits
Saving /mnt/extraspace/damonge/Datasets/KiDS1000/xcell_runs/KiDS1000_bin3_cat.fits
KiDS1000_bin4_cat.fits
Saving /mnt/extraspace/damonge/Datasets/KiDS1000/xcell_runs/KiDS1000_bin4_cat.fits
```

PR #268, opened 2023-08-15
Fix noise and rerun when removing overlap

### LINKED ISSUE #265: fix coupled_noise when removed overlap
When removing some overlap, we have to make sure that `w2s2` is 0 in the removed area.

carlosggarcia:
Closes #265, #266 and similar bugs. Fixes the noise and the catalog rerun paths when removing the overlap.

### DIFF
--- a/xcell/mappers/mapper_DELS.py
+++ b/xcell/mappers/mapper_DELS.py
@@ -85,7 +85,9 @@ class MapperDELS(MapperBase):
             catalog (Array)
         """
         if self.cat_data is None:
-            fn = f'{self.map_name}_cat.fits'
+            # This will handle the situation when removing the overlap
+            map_name = self.map_name.split("_")[0] + f"_bin{self.zbin}"
+            fn = f'{map_name}_cat.fits'
             self.cat_data = self._rerun_read_cycle(fn, 'FITSTable',
                                                    self._get_catalog)
         return self.cat_data
--- a/xcell/mappers/mapper_HSC_DR1wl.py
+++ b/xcell/mappers/mapper_HSC_DR1wl.py
@@ -118,7 +118,9 @@ class MapperHSCDR1wl(MapperBase):
             cat (Array)
         """
         if self.cat is None:
-            fn = f'{self.map_name}.fits'
+            # This will handle the situation when removing the overlap
+            map_name = self.map_name.split("_")[0] + f"_{self.bn}"
+            fn = f'{map_name}.fits'
             self.cat = self._rerun_read_cycle(fn, 'FITSTable',
                                               self._get_catalog_from_raw)
         return self.cat
--- a/xcell/mappers/mapper_HSC_DR1wl.py
+++ b/xcell/mappers/mapper_HSC_DR1wl.py
@@ -227,6 +227,8 @@ class MapperHSCDR1wl(MapperBase):
                                       cat[self.w_name]**2),
                                    ra_name='ra', dec_name='dec',
                                    rot=self.rot)
+        msk = self.get_mask()
+        w2s2[msk == 0] = 0
         return w2s2
 
     def get_nl_coupled(self):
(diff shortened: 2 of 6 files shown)


xcell/mappers/mapper_DELS.py
```diff
@@ -85,7 +85,9 @@ def get_catalog(self):
             catalog (Array)
         """
         if self.cat_data is None:
-            fn = f'{self.map_name}_cat.fits'
+            # This will handle the situation when removing the overlap
+            map_name = self.map_name.split("_")[0] + f"_bin{self.zbin}"
+            fn = f'{map_name}_cat.fits'
```
damonge, line 90: Not sure why you need to do this for DELS. This seems to be cutting it in redshift?

xcell/mappers/mapper_HSC_DR1wl.py
```diff
@@ -118,7 +118,9 @@ def get_catalog(self):
             cat (Array)
         """
         if self.cat is None:
-            fn = f'{self.map_name}.fits'
+            # This will handle the situation when removing the overlap
+            map_name = self.map_name.split("_")[0] + f"_{self.bn}"
+            fn = f'{map_name}.fits'
```
damonge, line 123: same question here as in DELS

xcell/mappers/mapper_KV450.py
```diff
@@ -70,7 +70,9 @@ def get_catalog(self):
             array: catalog data
         """
         if self.cat_data is None:
-            fn = f'{self.map_name}_cat.fits'
+            # This will handle the situation when removing the overlap
+            map_name = self.map_name.split("_")[0] + f"_bin{self.zbin}"
+            fn = f'{map_name}_cat.fits'
```
damonge, line 75: As above

xcell/mappers/mapper_KiDS1000.py
```diff
@@ -72,7 +72,9 @@ def get_catalog(self):
             cat_data (Array)
         """
         if self.cat_data is None:
-            fn = f'{self.map_name}_cat.fits'
+            # This will handle the situation when removing the overlap
+            map_name = self.map_name.split("_")[0] + f"_bin{self.zbin}"
+            fn = f'{map_name}_cat.fits'
```
damonge, line 77: same

xcell/mappers/mapper_WIxSC.py
```diff
@@ -105,7 +105,9 @@ def _get_catalog(self):
 
     def get_catalog(self):
         if self.cat_data is None:
-            fn = f'{self.map_name}_rerun_coord{self.coords}.fits'
+            # This will handle the situation when removing the overlap
+            map_name = self.map_name.split("_")[0] + f"_bin{self.zbin}"
+            fn = f'{map_name}_rerun_coord{self.coords}.fits'
```
damonge, line 110: and again

xcell/mappers/mapper_DELS.py
carlosggarcia: Yep, that's how it was before too. `self.map_name` was `DELS_binX`. So here I'm just keeping this structure by removing the part from the removed overlap. 

Same in the others

damonge: I don't understand. So `split("_")[0]` is meant to remove some part of the map name that denotes that you've cut some area?

carlosggarcia: That is correct. This is appended in `MapperBase` https://github.com/xC-ell/xCell/blob/9e6e4ff1e8c6a0fe2641dff5ae1a12400fa76406/xcell/mappers/mapper_base.py#L39-L43

damonge: oh, wow, that's very niche! OK then.

damonge (approved):

PR #272, opened 2023-11-09
DELS fixed

damonge:
Issue had to do with single vs. double precision, and an update to how either healpy or numpy deal with them.


JaimeRZP (commented):
LGTM!!

JaimeRZP (approved):

PR #274, opened 2023-11-21
Mapper base docs

JaimeRZP:


JaimeRZP:
@damonge Do you have a clue of why adding these strings are breaking the tests?

JaimeRZP:
it is not a flake8 issue

PR #275, opened 2024-02-09
Point source masking for ROSAT

damonge:


cosmotheka/mappers/mapper_ROSAT.py
```diff
@@ -47,17 +47,39 @@ def __init__(self, config):
...
+        self.nside_ps = 2048
+        self.count_map = None
         self.expmap = None
         self.pholist = None
         # signal_map is a map of the countrate
         self.nl_coupled = None
+        self.ps_mask = None
 
-    def get_pholist(self):
+    def _get_ps_mask(self):
```
carlosggarcia, line 66: Is this fast or should we save a temporal file?

cosmotheka/mappers/mapper_ROSAT.py
```diff
@@ -113,17 +157,18 @@ def _get_mask(self):
             msk = rotate_mask(msk, self.rot, binarize=True)
             msk = hp.ud_grade(msk, nside_out=self.nside)
             mask *= msk
+        if self.mask_ps:
+            # Mask point sources
+            ps_mask = hp.ud_grade(self._get_ps_mask(),
```
carlosggarcia, line 162: Duplicated of L.128? Maybe `ps_mask` can be stored to avoid changing the resolution twice?

cosmotheka/mappers/mapper_ROSAT.py
```diff
@@ -47,17 +47,39 @@ def __init__(self, config):
...
+            cat = cat[cat['RATE'] >
+                      self.config['point_source_flux_cut_ctps']]
+            nmap = get_map_from_points(cat, self.nside_ps,
+                                       ra_name='RA_DEG',
+                                       dec_name='DEC_DEG',
+                                       rot=self.rot)
+            self.ps_mask = (nmap == 0).astype(float)
+        return self.ps_mask
+
+    def _get_pholist(self):
```
carlosggarcia, line 82: Maybe we want to have the catalog methods public? Sth to discuss in all mappers, not only this one.

carlosggarcia (changes requested):
Minor comments/questions about optimization.

cosmotheka/mappers/mapper_ROSAT.py
damonge: it's just O(1000) sources. it's super fast.

cosmotheka/mappers/mapper_ROSAT.py
damonge: the less we make public, the less we have to document and maintain. But yes, we can discuss outside of this.

cosmotheka/mappers/mapper_ROSAT.py
damonge: No, we want to remove the point sources in the map at the native resolution (not at the potentially downgraded one). Remember it's the same thing we do for point sources in the SPT maps. It's also the reason why we want `_get_ps_mask` to return the mask in the native high resolution.

carlosggarcia: I understand that, but L128 is exactly the same this line, as far I can see

carlosggarcia: fine

cosmotheka/mappers/mapper_ROSAT.py
carlosggarcia: Yeah. I think we can think of a way of unifying all the catalog-based mappers so they share a public `get_catalog` method. That way we just document it once. 
Future work, anyway.

carlosggarcia:
LGTM

carlosggarcia (approved):

PR #281, opened 2024-04-09
Added readthedocs support

carlosggarcia:
What the title says


carlosggarcia:
The error has to do with pixell and cython. Unrelated to the changes I've done, so i'll merge and let it be

PR #282, opened 2024-04-11
Rebrand

carlosggarcia:
Updated some remaining `xcell`

We should decide what we do with the `xcell_rerun` in the input files.


damonge (approved):
LGTM

PR #284, opened 2024-04-16
Debug

carlosggarcia:
Fixed HSC and CatWISE mappers. 
 - For HSC, I have inverted the sign of the shear component e1. HCS x DECALS was also negative, so I think this is the right thing to do. 
 - For CatWISE, I have changed the way the ecliptic correction is applied: it is now subtracted, following the Cell 19 of https://github.com/rameez3333/CatWISEdipole/blob/main/CatWISE_Dipole_Results.ipynb and have made sure the number counts map is always positive. I am unsure why we have to multiply by the pixel area since by Cell 19 it seems that they obtain `rho_nc = fit * ecliptic_map / Area`, so the number count map should be `fit * ecliptic_map`.

### DIFF
--- a/cosmotheka/mappers/mapper_CatWISE.py
+++ b/cosmotheka/mappers/mapper_CatWISE.py
@@ -25,7 +25,7 @@ class MapperCatWISE(MapperBase):
         self.apply_ecliptic_correction = \
             config.get('apply_ecliptic_correction', True)
         self.cat_data = None
-        self.nmap_data = None
+        self.nmap_data = {'corr': None, 'uncorr': None}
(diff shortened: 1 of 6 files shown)


carlosggarcia:
After some tests with @damonge, we have found that:
 - HSC needed both e1, e2 to have the sign swapped
 - CatWISE noise was overestimated: one needed to use the uncorrected noise map when estimating the variance of the field.

carlosggarcia:
@damonge ready for your review :)

carlosggarcia:
David approves. Merging

PR #294, opened 2025-06-19
Include new mapper class for ACT DR6 tSZ maps

adrien-laposta:


cosmotheka/mappers/mapper_ACT_compsept.py
```diff
@@ -12,14 +12,22 @@ class MapperACTCompSept(MapperACTBase):
...
         # weight for each pixel corresponding to the detector array.
         signal_map = enmap.read_map(self.file_map)
-        signal_map = reproject.healpix_from_enmap(signal_map,
-                                                  lmax=self.lmax,
-                                                  nside=self.nside)
+        signal_map = reproject.map2healpix(
+            signal_map,
+            nside=self.nside,
+            lmax=self.lmax,
+            niter=3
```
carlosggarcia, line 29: Why `niter=3`? By default we use `niter=0` for the spherical harmonic transform. There were some effects with niter=3
https://github.com/Cosmotheka/Cosmotheka/blob/d66b83d28251a1323010399b44cfe76a85a57e35/input/master.yml#L83C1-L83C16

cosmotheka/mappers/mapper_ACT_compsept.py
```diff
@@ -12,14 +12,22 @@ class MapperACTCompSept(MapperACTBase):
     """
     def __init__(self, config):
         self._get_ACT_defaults(config)
+        if self.lmax > 3 * self.nside:
+            # WARNING
+            print("WARNING:              you selected lmax > 3*nside, "
```
carlosggarcia, line 17: Use `warnings.warn`

carlosggarcia (changes requested):
Thanks! Just a few comments/questions

cosmotheka/mappers/mapper_ACT_compsept.py
```diff
@@ -26,7 +26,6 @@ def _get_signal_map(self):
             signal_map,
             nside=self.nside,
             lmax=self.lmax,
-            niter=3
```
carlosggarcia: Not sure what this exactly does and what is the best option but maybe we can use here the same as for the `NmtField`? If so, I would read it from the config file. I think `self.config['sphere']['n_iter_sht']`

carlosggarcia (changes requested):
Last comment/question!

cosmotheka/mappers/mapper_ACT_compsept.py
carlosggarcia: sorry, i got confused. the `master.yml` has the `sphere` section, but for the mapper i think you should pass `n_iter_sht` directly, like for the nside and coords

PR #298, opened 2025-06-26
Include measured Nl for DR6 kappa

adrien-laposta:


damonge:
I added a test myself, but while I was at it I realised I have a question. See comment below

cosmotheka/mappers/mapper_ACTDR6k.py
```diff
@@ -64,8 +71,16 @@ def _get_mask(self):
...
-        # raise NotImplementedError("No noise model for the ACT maps")
-        return np.zeros([1, 3*self.nside])
+        if self.nl_coupled is None:
+            ell, nl = np.loadtxt(self.file_noise, unpack=True)
+            nl = interp1d(
+                ell, nl, bounds_error=False,
+                fill_value=(nl[0], nl[-1])
+            )(self.get_ell())
+            # Rescale to "couple" noise
+            nl *= np.mean(self.get_mask()**2)
```
damonge, line 81: The map is a CAR map, right? So if you want to correct for the coupling effect, shouldn't one take into account the different pixel sizes and the fact that the map is not full-sky when computing  <w^2>?

adrien-laposta: The map is not a CAR map. The release includes alms for lensing and a healpix mask so I guess this should not be an issue

PR #302, opened 2025-07-08
Desi lrg input

carlosggarcia:
Different additions/fixes to make a run with the LRGs and speed up the computation


carlosggarcia:
@damonge These are a mixture of changes i've done to make the runs easier and faster to run. There might be more changes coming in the future, but I think it's worth to move these to master for now, since not sure I will be able to implement the ideas i've got in the near future.

damonge (approved):

PR #306, opened 2026-02-24
NPipe Planck maps

damonge:
Work by @adrien-laposta


damonge (commented):
Cosmetic stuff. I think we can merge after fixing these things and get back to it if we see any issues.

PR #307, opened 2026-07-09
DESI BGS

damonge:
by @adrien-laposta


cosmotheka/mappers/mapper_DESI_BGS.py
```diff
@@ -0,0 +1,618 @@
...
+            boolmask = np.isin(spec_cat["TARGETID"], self.cat["TARGETID"])
+            zspec = spec_cat["Z"][boolmask]
+
+            bins = np.linspace(0, 1, 101)
+            nz, edges = np.histogram(zspec, bins=bins)
+
+            zmin, zmax = edges[:-1], edges[1:]
+            z_mid = zmin + (zmax - zmin) / 2
+
+            self.dndz = {"z_mid": z_mid, "nz": nz}
```
damonge: Since these things can take a while I would wrap this in a `read_rerun_cycle`

damonge: (and that way you'll also have the dndzs easily accessible in the future)

cosmotheka/cls/cov.py
```diff
@@ -178,25 +178,21 @@ def get_covariance_workspace(self, save_cw=True):
...
             self.outdir, f"cw__{mask1}__{mask2}__{mask3}__{mask4}.fits"
         )
-        cw = nmt.NmtCovarianceWorkspace()
         recompute = self.data.data["recompute"]["cmcm"]
         if (not recompute) and os.path.isfile(fname):
-            tools.read_wsp(cw, fname)
-            if cw.wsp is not None:
+            cw = tools.read_wsp(fname, True)
+            if cw is not None:
+                self.cw = cw
```
carlosggarcia, line 185: Good catch! Maybe this explains why my runs were kind of slow.

cosmotheka/cls/tools.py
```diff
@@ -67,13 +68,27 @@ def read_wsp(wsp, fname, **kwargs):
         Workspace or Covariance workspace to save
     fname: str
         Path to save the workspace to
+    cov: bool
+        If ``True``, you want to rea a covariance workspace.
```
carlosggarcia: "rea" -> "read"

cosmotheka/cls/tools.py
```diff
@@ -67,13 +68,27 @@ def read_wsp(wsp, fname, **kwargs):
...
+        w = cls.from_file(fname, **kwargs)
+        return w
+    except OSError as e:
+        if ('FITSIO' in str(e)) and os.path.isfile(fname):
+            print(f"Error reading {fname}. Removing it and computing it again")
+            os.remove(fname)
+            return
+
+        raise e
     except RuntimeError as e:
```
carlosggarcia, line 92: Is this now outdated?

cosmotheka/mappers/mapper_DESI_LRG.py
carlosggarcia, line 1: I guess you've checked you get the same as in the old way?

input/tracers_config/DESI_LRG__0.yml
```diff
@@ -1,16 +1,24 @@
...
-file_dndz: /mnt/extraspace/damonge/Datasets/DESI/lrg_xcorr_2023_v1/redshift_dist/main_lrg_pz_dndz_iron_v0.4_dz_0.02.txt
-stardens_path: /mnt/extraspace/damonge/Datasets/DESI/lrg_xcorr_2023_v1/misc/pixweight-dr7.1-0.22.0_stardens_64_ring.fits
-imaging_weights_coeffs: /mnt/extraspace/damonge/Datasets/DESI/lrg_xcorr_2023_v1/catalogs/imaging_weights/main_lrg_linear_coeffs_pz.yaml
+data_catalog: /mnt/extraspace/damonge/Datasets/DESI_LRG_PZ/dr9_lrg_pzbins.fits
+weights_catalog: /mnt/extraspace/damonge/Datasets/DESI_LRG_PZ/dr9_lrg_pzbins-weights.fits
+file_dndz: /mnt/extraspace/damonge/Datasets/DESI_LRG_PZ/main_lrg_pz_dndz_iron_v0.4_dz_0.02.txt
+
+external_maps:
+  - name: stardens
+    path: /mnt/extraspace/damonge/Datasets/GalacticMaps/stardens_gaia.fits
```
carlosggarcia, line 9: is it the same as the "official" one?

input/tracers_config/DESI_LRG__0.yml
```diff
@@ -1,16 +1,24 @@
...
+    threshold: 2500
+  - name: ebv
+    path: /mnt/extraspace/damonge/Datasets/GalacticMaps/lambda_sfd_ebv_Equatorial.fits
```
carlosggarcia, line 12: hmm, in the previous version it was not a map but a column in the catalog?

input/tracers_config/DESI_LRG__0.yml
carlosggarcia, line 1: The questions here apply to the other ymls too

input/tracers_config/DESI_LRG_Zhou23__0.yml
```diff
@@ -3,7 +3,19 @@ sample: main
 data_catalog: /mnt/extraspace/damonge/Datasets/DESI/lrg_xcorr_2023_v1/catalogs/dr9_lrg_pzbins.fits
 weights_catalog: /mnt/extraspace/damonge/Datasets/DESI/lrg_xcorr_2023_v1/catalogs/more/dr9_lrg_pzbins-weights.fits
 file_dndz: /mnt/extraspace/damonge/Datasets/DESI/lrg_xcorr_2023_v1/redshift_dist/main_lrg_pz_dndz_iron_v0.4_dz_0.02.txt
-stardens_path: /mnt/extraspace/damonge/Datasets/DESI/lrg_xcorr_2023_v1/misc/pixweight-dr7.1-0.22.0_stardens_64_ring.fits
+
```
carlosggarcia, line 6: I see, you keep the old files for this one. Then the question I've got is how much the change of maps affect the angular power spectra.

cosmotheka/tests/test_tools.py
```diff
@@ -103,19 +103,9 @@ def test_read_wsp(cwsp):
     else:
         assert w.wsp.lmax == w2.wsp.lmax
 
-    # Check you can pass kwargs
```
carlosggarcia, line 106: Why did you remove these? kwargs can still be passed, right?

cosmotheka/tests/test_cls.py
```diff
@@ -255,6 +255,7 @@ def test_cov_nlmarg():
 #       b) computation of missing terms. E.g. 1h is already in but the user
 #          requests 1h+2h later on. It only computes 2h and add it to the 1h
 #          contribution already there.
+@pytest.mark.skip("slow")
```
carlosggarcia: Can you add a comment saying we should run them manually? Or maybe it can be automatize somehow so that GitHub does it for us monthly or so.

cosmotheka/mappers/mapper_DESI_BGS.py
```diff
@@ -0,0 +1,607 @@
...
+        cuts = {
+            "target_maskbits": [1, 12, 13],
+            "min_nobs": 2,
+            "max_sigmaz": 0.05,
+            "remove_island": True,
+        }
+        return cuts
+
+    def _get_quality_cuts(self, cat, randoms=False):
+        """
```
carlosggarcia, line 70: Write the doc

cosmotheka/mappers/mapper_DESI_BGS.py
```diff
@@ -0,0 +1,607 @@
...
+        """
+        mask = np.ones(len(cat), dtype=bool)
+
+        # # Veto mask
+        # mask *= cat["lrg_mask"][:] == 0
+        # print("Veto mask. Keeping ", mask.sum())
+
+        # MASKBITS cut. The veto mask for randoms seem to miss some
```
carlosggarcia: This seems from the LRG mapper. I guess you want to remove the commented block from the lrg_mas and update this comment.

cosmotheka/mappers/mapper_DESI_BGS.py
```diff
@@ -0,0 +1,607 @@
...
+        return mask
+
+    def _load_spec_catalog(self):
+        """
+        """
+        cat = Table(fitsio.read(self.config["spec_catalog_xmatch"]))
+        # Restriction from 2510.14135 (don't know how relevant it is)
+        mask = cat["COADD_FIBERSTATUS"] == 0
+        mask &= np.isin(cat["ZWARN"], [0, 4])
+        mask &= np.isin(cat["SPECTYPE"], ["GALAXY", "QSO"])
```
carlosggarcia, line 120: hmm, if you want bgs, why do you select qso's?

cosmotheka/mappers/mapper_DESI_BGS.py
```diff
@@ -0,0 +1,607 @@
...
+    def _get_nz(self):
+        spec_cat = self._load_spec_catalog()
+        if self.cat is None:
+            self.get_catalog()
+
+        boolmask = np.isin(spec_cat["TARGETID"], self.cat["TARGETID"])
+        zspec = spec_cat["Z"][boolmask]
+
+        bins = np.linspace(0, 1, 101)
+        nz, edges = np.histogram(zspec, bins=bins)
```
carlosggarcia, line 135: do they all have spec-z? if so, why don't you do the binning based on the spec-z? if not, wouldn't you need to, at least, do DIR to reweight the n(z) distribution?

cosmotheka/mappers/mapper_DESI_BGS.py
```diff
@@ -0,0 +1,607 @@
...
+            self.dndz = self._rerun_read_cycle(fn, 'NPZ', self._get_nz)
+        return self._get_shifted_nz(dz)
+
+    def get_dtype(self):
+        return self.dtype
+
+    def get_spin(self):
+        return self.spin
+
+    def _get_list_randoms(self):
```
carlosggarcia: This method, the download randoms and maybe all the other randoms-related methods  are a copy of the LRGs code. Wouldn't make sense to make a parent class that was inherited by BGS, LRGs and any other new mapper we create to avoide repetion?

cosmotheka/mappers/mapper_DESI_BGS.py
```diff
@@ -0,0 +1,607 @@
...
+        except Exception as e:
+            print(f"Error downloading {url}: {e}")
+            raise
+
+        return rand_file
+
+    def _load_full_randoms(self, base_name):
+        random_path = self._randoms_path
+        rand_file = os.path.join(random_path, f"{base_name}.fits")
+        rand_mask_name = f"{base_name}-lrgmask_v1.1.fits.gz"
```
carlosggarcia: can you use the lrgmask?

cosmotheka/mappers/mapper_DESI_BGS.py
```diff
@@ -0,0 +1,607 @@
...
+    def get_randoms_maps(self):
+        if self.randoms_maps["n"] is not None:
+            return self.randoms_maps
+
+        list_randoms = self._get_list_randoms()
+
+        randoms_maps = np.zeros((3, self.npix))
+
+        # Hack to remove the density definition from the randoms map name
+        map_name = self.map_name.replace("_densdefZhou2023", "")
```
carlosggarcia: does this make sense in the class?

cosmotheka/mappers/mapper_DESI_BGS.py
```diff
@@ -0,0 +1,607 @@
...
+        return self.randoms_maps
+
+    def compute_weights(self, randoms):
+        """
+        Compute the weights for the randoms.
+        :param randoms: astropy Table with the randoms
+        :return: weights per z-bin
+        """
+        # Placeholder, need to check where data products are
+        return {"weight_pzbin": np.ones(len(randoms))}
```
carlosggarcia, line 174: are you happy to leave this like this?

cosmotheka/mappers/mapper_DESI_BGS.py
carlosggarcia, line 1: I guess you've checked that what you get makes sense? I see many repetitions wrt the LRG mapper. Maybe it'd be possible to create a parent class to avoid repetition and prepare the ground for, maybe, QSOs? 

I've spotted some lines that might be LRG specific. See comments below.

input/tracers_config/DESI_BGS__0.yml
```diff
@@ -0,0 +1,30 @@
...
+    path: /mnt/extraspace/damonge/Datasets/GalacticMaps/lambda_sfd_ebv_Equatorial.fits
+    threshold: 0.15
+
+spec_catalog_xmatch: /mnt/extraspace/damonge/Datasets/DESI_DR1/datcomb_BGS_BRIGHT_tarspecwdup_zdone.fits
+
+
+# RANDOMS
+randoms_path: /mnt/extraspace/damonge/Datasets/DESI/randoms_for_lrg_xcorr_2023_v1
+randoms_selection: /mnt/extraspace/damonge/Datasets/DESI/randoms_for_lrg_xcorr_2023_v1/randoms_Rongpu.txt
+# randoms_selection: /mnt/extraspace/alaposta/BGS_reruns/randoms_Rongpu.txt
```
carlosggarcia: remove old comment

carlosggarcia (changes requested):
Few comments and questions. The main issues are:
- There is no test file for the BGS
- There is a lot of repetition between BGS and LRG. I'd suggest creating a parent class.
- There are a few things in the BGS that I am not sure how good they are: using the same mask as lrg's, not calibrating the n(z) but using the spec-z directly, and few other comments that you'll see in the review.
- There are some changes in the LRGs that not sure if they will impact the final Cell. It took me some time to get something sensible there. Have you checked these changes don't screw the results?

cosmotheka/cls/tools.py
damonge: no, we need both. Basically covariance workspaces and normal workspaces throw different errors (for now, the newer version of NaMaster, which is still not released, will homogenise how wsps are read)

input/tracers_config/DESI_LRG__0.yml
damonge: it's exactly the same, just saved as an actual healpix map

input/tracers_config/DESI_LRG__0.yml
damonge: yes, which was just the map evaluated at the position of the sources. This has the advantage that we can now use better versions of the EBV maps that came out after the old SFD map.

input/tracers_config/DESI_LRG__0.yml
damonge: Same answers

input/tracers_config/DESI_LRG_Zhou23__0.yml
damonge: nothing changes, this just allows us to homogenise how we deal with BGS and LRG.

Incidentally, do we really want to kep the Zhou version of the mapper around? It seems a nuisance.

cosmotheka/tests/test_tools.py
damonge: No, `read_unbinned_MCM` is no longer supported in NaMaster. This is completely removed in the next PR.

cosmotheka/mappers/mapper_DESI_BGS.py
damonge: @adrien-laposta may be able to clarify, but I guess some of the photometric BGS could be misclassified QSOs. Note that the cross-matching happening in the next method is done by TARGETID, so if there are objects classified as QSOs in the photometric BGS sample, we want to take them into account!

cosmotheka/mappers/mapper_DESI_BGS.py
damonge: the spectroscopic coverage is very large but not 100% complete. The colour matching is essentially exact, so no need to DIR.

cosmotheka/mappers/mapper_DESI_BGS.py
damonge: Yes, will do.

cosmotheka/mappers/mapper_DESI_BGS.py
damonge: comment was wrong. We've checked that we don't need weights for BGS (essentially systematics-free)

cosmotheka/mappers/mapper_DESI_BGS.py
damonge: Yes, will make a parent class and report back.

input/tracers_config/DESI_BGS__0.yml
damonge: done

cosmotheka/mappers/mapper_DESI_BGS.py
damonge: yes, removed

cosmotheka/mappers/mapper_DESI_BGS.py
damonge: done

cosmotheka/tests/test_cls.py
damonge: Added comment

cosmotheka/cls/tools.py
damonge: Done

input/tracers_config/DESI_LRG__0.yml
carlosggarcia: Got it. thanks!

input/tracers_config/DESI_LRG_Zhou23__0.yml
carlosggarcia: I'd keep it at least until we finish the project with Matteo, in case we have to show comparison (remember there is a small difference due to the way `delta` is defined.

cosmotheka/tests/test_tools.py
carlosggarcia: and aren't there any other kwargs that can be passed?

cosmotheka/mappers/mapper_DESI_BGS.py
carlosggarcia: hmm, if, imagine, there are less high-z with spec-z, wouldn't your n(z) be biased low?

cosmotheka/mappers/mapper_DESI_BGS.py
damonge: we are not imposing the lrg mask, this is just downloading / loading that column, but we never use it. This is gone in the newest commit, where the BGS mapper is a subclass of the LRG one. Note, in `_get_quality_cuts` that we never use the lrgmask column

cosmotheka/tests/test_tools.py
damonge: nope...

cosmotheka/mappers/mapper_DESI_BGS.py
damonge: But this is not the case, right? The photometric BGS sample has been followed up almost entirely by DESI. Basically, in the patches that have been observed, all of these galaxies have spectra.

cosmotheka/mappers/mapper_DESI_BGS.py
damonge: BGS now inherits from LRG and I've eliminated as much repeated code as was reasonable.

cosmotheka/mappers/mapper_DESI_LRG.py
damonge: Yes, the student working with us checked that he gets the same results as in Adrien's previous paper with this new mapper.

damonge:
@carlosggarcia thanks a lot for these. On the global comments:

- I've now subclassed BGS from the LRGs, removing a lot of repeated code. Definitely needed to do that.
- Yes, we've checked that the reshuffling we've done in the LRG code doesn't impact the final spectra.
- See above about the BGS N(z). The point is that there is almost exact spectroscopic coverage.

I'll write unit tests for the BGS mapper, but in the meantime you may want to look at the latest changes. I'll ping you again when the tests are ready.

cosmotheka/mappers/mapper_DESI_BGS.py
damonge: No longer there

damonge:
OK, all done including unit tests @carlosggarcia . Back to you

cosmotheka/mappers/mapper_DESI_LRG.py
carlosggarcia: sounds good. thanks

cosmotheka/mappers/mapper_DESI_BGS.py
carlosggarcia: ah, I see. I wasn't aware of how the survey strategy is. I thought it may be like Rubin, and go wide first and deep after.

cosmotheka/mappers/mapper_DESI_BGS.py
carlosggarcia: not ideal but we can live with this for now until we implement a new sample

cosmotheka/mappers/mapper_DESI_BGS.py
```diff
@@ -1,87 +1,29 @@
 import numpy as np
-import healpy as hp
 import fitsio
-from .mapper_base import MapperBase
+from .mapper_DESI_LRG import MapperDESILRG
```
carlosggarcia, line 3: My only worry of inheriting from LRGs directly is that something that should be redefined in BGS is not and then it is used for the BGS inadvertently.

cosmotheka/mappers/mapper_DESI_BGS.py
adrien-laposta: About this, I linked the paper above in the code, it is just to mimic the treatment in Boryana's paper (bottom of page 4). Agree with David above, might just be for misclassified galaxies

cosmotheka/mappers/mapper_DESI_LRG.py
```diff
@@ -554,10 +559,13 @@ def compute_weights(self, randoms):
     def _load_full_randoms(self, base_name):
         random_path = self._randoms_path
         rand_file = os.path.join(random_path, f"{base_name}.fits")
-        rand_mask_name = f"{base_name}-lrgmask_v1.1.fits.gz"
-        lrgmask_file = os.path.join(
-            self.config["randoms_lrgmask_path"], rand_mask_name
-        )
+        if "randoms_lrgmask_path" in self.config:
```
carlosggarcia, line 562: This is also the problem of using the LRGs as parent class.

carlosggarcia (changes requested):
A few comments. I'd suggest not inheriting from the LRGs but I will let you up to your choice.

Also, the tests for the BGS have not been submitted?

cosmotheka/mappers/mapper_DESI_BGS.py
damonge: there's nothing like that currently (BGS just require fewer things than LRGs)

cosmotheka/mappers/mapper_DESI_LRG.py
damonge: but I'd need this regardless of which class I subclass from. The other option is to repeat a lot of code (except for this if statement) in both subclasses.

damonge:
Ooops. Tests now added. Sorry about that!

carlosggarcia (approved):
LGTM

PR #310, opened 2026-08-01
Fix MC correction cache

carlosggarcia:
This PR fixes the MC correction cache and modifies the `run_cls_mpi.py` so it can be run without MPI (e.g. when you want to just create the sacc file, because that is done in a single job).

@damonge let me know if you want me to remove the old script `run_cls.py`.


damonge (approved):
LGTM. Let's leave run_cls, since I still use it sometimes :-)

